Issue #333, opened 2026-07-05
Knowledge layer: AGENTS.md block + Cursor rule + Claude Code Skill for dataprof


AndreaBozzo:
## Context

MCP gives agents *connectivity*; this issue gives them *knowledge* — teaching Claude Code, Codex, and Cursor how to actually use dataprof well. `AGENTS.md` is the cross-tool standard (read natively by Codex and Cursor; Claude Code reads it too but prefers Skills/CLAUDE.md).

## Deliverables (short, copy-pasteable)

- **`AGENTS.md` block** (a snippet users drop into their repo): teaches the workflow —
  *"To understand a dataset: call `dp.analyze_structure()` first for a cheap look, then `dp.profile()` for full metrics. Use `report.to_llm_context()` for a compact summary, `report.compare(other)` for before/after drift. Don't dump raw rows."*
- **Cursor rule** (`.cursor/rules/dataprof.md`) — same guidance in Cursor's format.
- **Claude Code Skill** (`SKILL.md`) — a "profile this dataset" skill packaging the workflow as on-demand knowledge.

Keep each focused (the research is clear: a tight 50-line file outperforms a sprawling 500-line one).

## Acceptance
A user can copy any one of the three into their project and their agent immediately follows the intended dataprof workflow (structure → profile → interpret → export/compare) without dumping raw data.

Pairs with: MCP server (#330), docs page.

Issue #326, opened 2026-07-05
Python 0.9 breaking pass: DataQualityMetrics flat→nested semantics + requires-python


AndreaBozzo:
## Context

Split out from #315 because these two items are **semver-sensitive** and deserve deliberate review rather than being buried in a grab-bag. 0.9.0 (still pre-1.0) is the right window for one intentional breaking pass — but it must be explicit, documented, and tested.

### 1. `DataQualityMetrics` flat-vs-nested semantics
Flat accessors (`q.missing_values_ratio`, etc.) return `0.0` whether a dimension scored 100% **or was never computed** — a silent footgun. Decide and implement one of:
- **(preferred)** deprecate the flat accessors in favor of the nested dimension properties (`q.completeness`, …), which are `None` when a dimension wasn't requested; or
- have flat accessors raise when the dimension wasn't computed.

Pair with **#244** (table-driven ISO metric tests) so the chosen semantics are locked by tests for at least completeness / uniqueness / accuracy.

### 2. `requires-python`
`pyproject.toml` says `>=3.8`, but 3.9 is EOL and the stubs use PEP 604 unions (`str | int | None`) that don't parse at runtime on 3.8. Decide: bump to `>=3.10` (clean) or be deliberate about which annotations stay stub-only.

## Requirements
- CHANGELOG entry + short migration note for both changes.
- Tests covering the new semantics and the minimum supported Python.

## Out of scope
The non-breaking #315 items (discoverability, `list_patterns`, namespace cleanups, consistency fixes) stay in #315.

Issue #325, opened 2026-07-05
Python: ProfileReport.load(path) — reload a saved report from a file


AndreaBozzo:
## Context

#323 added `ProfileReport.from_dict()` / `from_json()` (read-only reload). But `from_json()` takes a JSON **string**, while the natural "reload a report I saved yesterday" flow starts from a **file path**:

```python
before.save("before.json")
loaded = dp.ProfileReport.from_json("before.json")   # ❌ tries to parse the filename as JSON
```

This is also the exact 0.9.0 acceptance snippet, so it must work end-to-end.

## Proposal

Add a `load()` classmethod symmetric with `save()`:

```python
@classmethod
def load(cls, path: str | os.PathLike) -> ProfileReport: ...
```

- `.json` → `from_json(file contents)`
- `.csv` / `.parquet` → out of scope for reload (these only store column profiles, not a full report) — raise a clear `ValueError` pointing users to `.json`.

Keep `from_json`/`from_dict` as the string/dict entry points; `load()` is the path entry point.

## Acceptance

```python
before = dp.profile("raw.csv")
before.save("before.json")
loaded = dp.ProfileReport.load("before.json")
assert loaded.quality_score == before.quality_score
```

Plus stub + docs + test. Part of the 0.9.0 "export → reload → compare" story.

JHON12091986:
Hi, I would like to work on this issue. Please assign it to me.

Thanks.

AndreaBozzo:
> Hi, I would like to work on this issue. Please assign it to me.
> 
> Thanks.

Hello @JHON12091986 and welcome to dataprof repo! Issue is now assigned to you as requested

Issue #319, opened 2026-05-31
Expand partial analysis into a configurable lightweight structural profiling surface


AndreaBozzo:
## Summary

`dataprof` already has a public partial-analysis surface, but today it is still limited to:

- `infer_schema*`
- `quick_row_count*`

That was the right first step, but it leaves a large gap between:
- very cheap structural checks, and
- full `profile()` / `ProfileReport`

This issue tracks the next step: turning partial analysis into a real, configurable, lightweight structural profiling surface.

## Why this matters

Many consumers do not want a full profile immediately. They want a cheap first pass that can answer questions like:

- What columns / fields are present?
- Roughly how many rows are there?
- Are there obvious null-heavy or sparse columns?
- Which columns look numeric, categorical, identifiers, or free text?
- What does the data shape look like before paying for full metrics / patterns / quality?

Today, `dataprof` computes some of the low-cost building blocks internally, but the public partial-analysis API exposes only a small subset of them.

## Current state

Current public partial surface:
- `infer_schema`
- `quick_row_count`
- async variants for streams

Current result types:
- `SchemaResult`
- `RowCountEstimate`

Current limitations:
- partial analysis is not configurable enough
- heuristics are hard-coded
- result types are too narrow for a real structural pass
- sync / async / Python parity is incomplete
- there is no lightweight `partial profile` report between `quick_row_count()` and full `profile()`

## Design goal

Define partial analysis as a cheap structural pass rather than only `schema + row count`.

This should stay meaningfully cheaper than full profiling, while surfacing more actionable information for users and downstream systems.

### 1. Add `PartialAnalysisConfig`

Introduce a dedicated config type for partial analysis, for example:

```rust
pub struct PartialAnalysisConfig {
    pub sample_rows: Option<usize>,
    pub row_count_sample_lines: Option<usize>,
    pub full_scan_threshold_bytes: Option<u64>,
    pub has_header: Option<bool>,
    pub csv_delimiter: Option<u8>,
    pub csv_flexible: Option<bool>,
    pub sample_value_capacity: Option<usize>,
    pub format_override: Option<FileFormat>,
}
```

Goals:
- remove hard-coded-only behavior
- let users control cost / accuracy tradeoffs
- make partial-analysis behavior more explicit and testable

### 2. Add a lightweight structural summary API

Add a richer API, e.g. one of:

- `partial_profile(...)`
- `quick_structural_summary(...)`
- `analyze_structure(...)`

[...]

PR #338, opened 2026-07-05
docs: add release notes and migration guide for 0.9.0

### LINKED ISSUE #329: 0.9.0 release notes + migration guide
## Context

0.9.0 introduces at least one breaking change (DataQualityMetrics semantics, #326) and a `requires-python` bump. Even a small break must be communicated clearly, or it erodes trust. This issue tracks a concise release-notes + migration doc — short, but it must exist.

## Contents

- **Breaking changes**
  - `DataQualityMetrics` flat→nested semantics (#326), incl. any deprecation of flat accessors.
  - `requires-python` change (#326).
- **What works in the base wheel** vs **what needs a source build** (async URL, database) — mirrors the README funnel (#327).
- **New this release**: `ProfileReport.load()` (#325), report `to_html`/`to_markdown`/`compare`/`from_json` (shipped in #323), minimal `analyze_structure()` (#319 first slice), curated real-world examples (#67).
- **Migration snippets** for anyone on the flat DataQualityMetrics accessors.

## Acceptance
A user upgrading to 0.9.0 can read one page and know: what broke, what to change, what the wheel gives them out of the box, and what's new. Keep it tight.

JHON12091986:
## Summary

This pull request adds comprehensive release notes and a migration guide for version 0.9.0, ensuring users can understand breaking changes and adopt new features smoothly.

## Type of Change

- Documentation update

## Changes Made

- Created `docs/release-notes.md` containing:
  - Breaking changes: DataQualityMetrics flat to nested semantics, Python version requirement update to 3.9+
  - New features: ProfileReport.load(), to_html(), to_markdown(), compare(), analyze_structure()
  - Migration guide with code examples for upgrading from 0.8.x
  - Full changelog referencing related issues

## Testing

- Documentation only, no code changes to test

## Documentation

- Release notes and migration guide added
- Examples included for each breaking change and new feature
- Clear upgrade path provided for existing users

## Checklist

- Documentation is complete and clear
- Examples are accurate and tested
- Migration path is fully explained

Closes #329

### DIFF
--- a/docs/release-notes.md
+++ b/docs/release-notes.md
@@ -1,0 +1,24 @@
+\# DataProf 0.9.0 Release Notes
+
+
+
+\## Breaking Changes
+
+
+
+\### DataQualityMetrics Semantics
+
+The `DataQualityMetrics` class has changed from flat to nested structure.
+
+
+
+\*\*Before (0.8.x):\*\*
+
+```python
+
+metrics = profile.quality\_metrics()
+
+print(metrics.completeness)
+
+print(metrics.uniqueness)
+


AndreaBozzo:
Thanks for this, and for picking up #329.

I'm going to keep this open and parked rather than merge it yet, and here is the honest reason: 0.9.0 is not built. Most of what these notes describe is still in open issues, not merged code. `ProfileReport.load()` (#325), `analyze_structure()` (#319), and the `DataQualityMetrics` flat-to-nested change (#326) have not landed, and the breaking decisions are not final. As one example, the minimum Python version in #326 is still an open choice between 3.9 and 3.10, so pinning "3.9+" here now could turn out wrong.

Release notes and a migration guide have to describe what actually shipped. If we merge them before the release exists, they drift from reality before 0.9.0 is even out, and users end up trusting a document that is partly fiction.

So this stays open, on hold, and we come back to it once 0.9.0 is feature complete. At that point it is exactly what #329 needs, and your draft is a good starting point.

If you would like something that can merge sooner, a few good entry points:

* #325 `ProfileReport.load()` (good first issue, small, well specified)
* one of the three real world examples in #67
* the AGENTS.md / Cursor rule / Claude skill snippets in #333

Appreciate the contribution. I don't want it lost, just timed correctly.

AndreaBozzo:
0.9.0 is out, so this can come off hold — and the reason I parked it turned out to be the right call, just not for the reason I gave.

When I wrote that comment, `ProfileReport.load()` (#325), `analyze_structure()` (#319) and the `DataQualityMetrics` change (#326) were all still open. They have since landed, and the Python floor question resolved to **3.10**, not the 3.9 this draft states. That single line would have shipped wrong.

More importantly, the breaking change described here didn't happen the way any of us expected. The flat `DataQualityMetrics` accessors were **deprecated, not removed** — they still return their 0.8-compatible values for the whole 0.9 line and merely emit a `DeprecationWarning`. So 0.9.0 is far less breaking than this draft says, and the migration guide's job turned out to be different: it only has to move people who run with `-W error`.

I've written the notes against what actually merged, and closed #329. Rather than rewrite this branch into something unrecognisable, I'm closing the PR — but your outline is what the final document is built on, and you're credited in its acknowledgements.

Two things worth knowing, because they're the kind of thing that's invisible until you go looking:

1. Every code snippet in the final notes was **executed against the built wheel** before publishing. That caught four errors in my own first draft, including that `report.quality` is a property rather than a method, and that the metrics are percentages on a 0-100 scale rather than ratios on 0-1. A migration guide that doesn't run is a migration guide that lies.
2. The file in this PR is backslash-escaped throughout (`\#`, `\*\*`, `quality\_metrics()`), so it renders as literal backslashes. Something in your editor escaped the markdown on save — worth checking, since it'll bite the next PR too.

Thank you for picking up a `help wanted` issue and for being patient while it sat. If you want another, #306 (per-crate READMEs) and #356 (contributor setup smoke test) are both in the 0.10.0 milestone, both `good first issue`, and neither depends on unreleased code — which is exactly what went wrong here.

PR #345, opened 2026-07-07
chore(python): adopt ty type checking with a compiled-core stub

AndreaBozzo:
## What

Adopts [ty](https://github.com/astral-sh/ty) (Astral's type checker) for the Python package, and gives it something real to check by stubbing the compiled core.

- **New `python/dataprof/_dataprof.pyi`** — declares the raw PyO3 extension symbols. Previously `dataprof._dataprof` was unresolvable to any type checker (mypy/pyright/ty alike); now it resolves. The public `__init__.pyi` re-exports these types instead of duplicating them, and gains the missing `ProfileReport.__init__`.
- **ty wired into CI as advisory** (non-blocking, since ty is pre-1.0) in the existing `Python Lint & Format` job; added to the `dev` dependency group + a `[tool.ty]` config block.

## Real findings fixed along the way

- `_df_config()` returns `ProfilerConfig | None` — stub params made nullable to match.
- `profile(on_progress=...)` was typed `object`; now `Callable[[ProgressEvent], None] | None`.
- Tests: narrow `report.quality` (an `Optional`) with asserts before attribute access.
- Move `typing.Iterator/Callable` → `collections.abc` (UP035, newly active under the py310 target).
- ruff: enable isort `combine-as-imports` for tidy re-export blocks.

## Verification

`uv run ty check python/` → **All checks passed**; `ruff check`/`format` clean; `pytest` **139 passed**.

Net result: the compiled extension is finally typed, which improves IDE/type-checker UX for downstream users too — independent of which checker they run.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -171,6 +171,11 @@ jobs:
       - name: Lint
         run: uv run ruff check python/
 
+      # Advisory: ty is pre-1.0 (Astral's type checker). Non-blocking for now.
+      - name: Type check (ty)
+        continue-on-error: true
+        run: uv run ty check python/
+
   python-tests:
     name: Python Tests
     runs-on: ubuntu-latest
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -7,6 +7,17 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Python: type stubs for the compiled extension (`dataprof/_dataprof.pyi`), so
+  type checkers and IDEs can resolve the native core. The public stubs now
+  re-export these types instead of duplicating them.
+
+### Internal
+
+- Python: adopt [ty](https://github.com/astral-sh/ty) for static type checking,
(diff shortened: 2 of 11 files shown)


AndreaBozzo:
Thanks — addressed both comments in 20915a2. Rather than swap `# ty: ignore` for `# type: ignore` (which trades one tool-specific form for another, since the error *codes* differ between ty and mypy), I removed both suppressions with fixes that work for every checker:

- **`__init__.py:620`** — the real cause was `_STOP_SHORTHANDS` being typed `dict[str, object]`; `object` genuinely isn't callable. Retyped it `dict[str, Callable[[], StopCondition]]`, so the value is callable with no suppression. (Note: `# type: ignore` here would actually be flagged as an *unused* ignore by mypy/pyright, since they narrow the guard fine — the annotation fix is the correct answer.)
- **`from_dict` (was :1254)** — used `typing.cast("_RustProfileReport", _DictBackedReport(data))` to express the intentional duck-typing. `cast` is honored by ty, mypy, and pyright alike.

No `# ty: ignore` remain in the tree; `ty check`, `ruff`, and `pytest` (139) all pass.

PR #369, opened 2026-07-10
test: add cross-engine parity suite

### LINKED ISSUE #363: Cross-engine parity tests: the same data must profile identically through every engine
## Problem

0.9.0 shipped two independent fixes to the same class of defect, in engines that are supposed to agree with each other:

- **#358** — Parquet/Arrow folded the physical value under a null (`0.0`) into `min`, `mean`, `std_dev` and `unique_count`, because `impl<T> From<T> for Option<T>` made the null guard always-true. Every nullable numeric column read through Parquet, pandas, polars or Arrow was wrong.
- **Database connectors** — every non-text column profiled as an all-null string, because a decode error was discarded with `.ok()` and became indistinguishable from SQL NULL.

Both survived a full test suite. Both were found the same way: **profile the same data through two engines and notice the numbers disagree.**

For the database fix this was the entire diagnostic:

```
sqlite  amt   type=string   nulls=3  unique=0  mean=None
csv     amt   type=float    nulls=1  unique=2  mean=20.0     <- same rows
```

Nothing in CI does this. Each engine is tested against its own hand-written expectations, so an engine can be confidently, consistently wrong.

## Proposal

A cross-engine parity suite. One fixture, materialised into every input format we support, profiled through every engine, asserting the resulting `ColumnProfile` values are identical.

Engines to cover: CSV, JSON/JSONL, Parquet, Arrow `RecordBatch` (PyCapsule), pandas, polars, dict-of-columns, list-of-row-dicts, and — behind the `database` feature — SQLite.

The fixture must include the cases that actually broke:

- a nullable numeric column with nulls *interleaved between* real values, not trailing (`[100.0, null, 1.0, null, 2.0]`)
- an integral float column (`100.0, 1.0, 2.0`) — must infer `float`, not `integer`
- an integer column containing a null — must stay `integer`, not widen to `float`
- an all-null column
- an integer beyond `2^53` — must not round-trip through `f64`
- a boolean column — must not render as `"true"`/`"false"` where an integer is expected
- a text column containing the literal string `"NULL"` and one containing `""`

Assert on the full profile, not a sampled field: `data_type`, `null_count`, `unique_count`, `min`, `max`, `mean`, `std_dev`.

## Acceptance

- A single parametrised test that fails if any two engines disagree on any field.
- Reverting either #358 or the database decoding fix makes it fail.
- Runs in CI on the default feature set; the database and polars arms skip cleanly when those features/packages are absent, in the style of the existing `importorskip` guards.

## Notes

Where engines *legitimately* differ (Parquet carries a real schema, CSV infers one), encode the difference as an explicit expected exception with a comment, rather than weakening the assertion. A parity suite whose exceptions list grows without justification is how this bug class comes back.

AndreaBozzo:
## Summary

- adds a cross-engine profile parity suite covering CSV, JSON, JSONL, Python mappings/rows, Parquet, Arrow, pandas, polars, and SQLite when database features are enabled
- fixes Arrow `NullArray` handling so all-null columns report their logical null count without fabricated string values
- runs the parity suite in the database CI build, enabling the SQLite arm there

## Why

Individual engine tests can all pass while an input path is consistently wrong. The canonical fixture encodes previously shipped nullable-Parquet and database-decoding regressions, and asserts each observable profile field against the direct core path.

## Validation

- `uv run pytest python/tests/test_engine_parity.py -ra`
- `uv run ruff format python/ .github/scripts/`
- `uv run ruff check python/ .github/scripts/`
- `cargo fmt --all -- --check`
- `cargo clippy --all --all-targets -- -D warnings`
- `cargo test -p dataprof-parquet --all-features test_all_null_column_profiles_as_all_nulls`

Closes #363

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -446,4 +446,4 @@ jobs:
         run: uv run maturin develop --features "python,python-async,database,sqlite"
 
       - name: Run Python database tests
-        run: uv run pytest python/tests/test_database_api.py -v -ra
+        run: uv run pytest python/tests/test_database_api.py python/tests/test_engine_parity.py -v -ra
--- a/crates/dataprof-parquet/src/arrow_profiler.rs
+++ b/crates/dataprof-parquet/src/arrow_profiler.rs
@@ -295,9 +295,16 @@ impl ColumnAnalyzer {
 
     fn process_array(&mut self, array: &dyn Array) -> Result<(), DataProfilerError> {
         self.total_count += array.len();
-        self.null_count += array.null_count();
+        // logical_null_count, not null_count: a NullArray has no validity
+        // buffer, so the physical count reports 0 nulls for an all-null array.
+        self.null_count += array.logical_null_count();
 
         match array.data_type() {
+            arrow::datatypes::DataType::Null => {
+                // Every slot of a NullArray is null, but is_null() is false on
+                // all of them (no validity buffer), so the string fallback
+                // below would fabricate one value per null slot.
+            }
             arrow::datatypes::DataType::Float64 => {
                 if let Some(float_array) = array.as_any().downcast_ref::<Float64Array>() {
                     self.process_float64_array(float_array)?;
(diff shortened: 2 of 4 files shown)


AndreaBozzo:
Addressed Copilot review: added a focused ArrowProfiler NullArray regression test (562ea66). The CSV parity arm was re-run and passes; because the fixture is emitted via Python csv.writer, its quoted empty string is preserved by the current parser, so no CSV exception is needed.

PR #370, opened 2026-07-10
fix(decode): audit swallowed errors on decode paths

### LINKED ISSUE #364: Audit swallowed errors: .ok() and unwrap_or_default() on decode paths
## Problem

The same mistake has now been found three times in this codebase: **an error is discarded and silently becomes "missing data."** A profiler that reports absence instead of failure is worse than one that crashes, because the wrong number is indistinguishable from a right one.

Confirmed instances:

1. **#358** (Parquet/Arrow) — `if let Some(value) = array.value(index).into()` never rejected anything, because `impl<T> From<T> for Option<T>` makes that arm always taken. The value under a null was folded into the statistics.
2. **Database connectors** — `let value: Option<String> = row.try_get(i).ok();` turned a sqlx *decode error* into `None`, which the profiler recorded as SQL NULL. Every non-text column, on all three connectors.
3. **#359** (agent context redaction) — `patterns` was a `Vec`, so "detection never ran" and "detection ran and found nothing" were the same empty value. The redaction gate read absence of evidence as evidence of absence, and printed raw credit-card numbers.

The third one is the general lesson: these are all cases where a type could not distinguish *unknown* from *empty*, and the code chose the reassuring interpretation.

## Proposal

Audit the codebase for the shape, then make it hard to reintroduce.

**Audit.** Every `.ok()`, `.unwrap_or_default()`, `.unwrap_or(0)`, `unwrap_or(0.0)` and `.ok().flatten()` on a path that produces a value the profiler will treat as data. For each, decide explicitly:

- the error is impossible → `expect()` with the reason
- the error means "no data" → keep, and say so in a comment
- the error means "we do not know" → propagate it, or represent it as `Option`/`None` distinct from the empty case

Start with the decode boundaries, which is where all three bugs lived: `crates/dataprof-parquet`, `crates/dataprof-db/src/connectors`, `crates/dataprof-python/src/arrow_export.rs`, and the CSV type inference.

**Guard.** Consider `#![deny(clippy::let_underscore_must_use)]` plus a narrow CI grep that fails on new `.ok()` immediately followed by `unwrap_or_default()` in the connector and decode modules. A lint that only covers the modules where this bug class lives is worth more than a global one everyone allows away.

## Acceptance

- Each surviving `.ok()` / `unwrap_or_default()` on a decode path carries a one-line comment naming which of the three cases it is.
- A regression test for any live bug the audit turns up.
- CI fails on a newly introduced instance in the audited modules.

## Related

- #363 cross-engine parity tests — catches the *consequence*; this issue attacks the *cause*.

AndreaBozzo:
## Summary

- audits decode paths and removes silent fallback behavior that could turn malformed input into plausible profile values
- preserves errors where conversion can fail, and documents the two in-memory serialisation cases that are intentionally infallible
- applies the required Rust formatter output

## Why

For a data profiler, a silent decode error is worse than a reported failure: it can produce a credible but incorrect statistic. This audit ensures decode paths surface errors instead of folding them into default values.

## Validation

- `cargo fmt --all -- --check`
- `cargo clippy --all --all-targets -- -D warnings`

Closes #364

### DIFF
--- a/.github/scripts/check_decode_audit.py
+++ b/.github/scripts/check_decode_audit.py
@@ -1,0 +1,92 @@
+"""Guard against silently swallowed errors on decode paths (issue #364).
+
+Every `.ok()`, `.unwrap_or_default()` or `.unwrap_or(0...)` in the audited
+decode modules must carry a `decode-audit:` classification comment on the same
+line or within the preceding lines, naming the case it falls into:
+
+- ``impossible`` — the error cannot happen (prefer ``expect()`` with a reason)
+- ``no-data``    — the absent value genuinely means "no data", not an error
+- ``unknown``    — the failure is surfaced (logged/propagated), never a default
+
+A new unclassified instance in these modules fails CI. See AGENTS.md
+("Error handling") and issue #364 for the rationale.
+"""
+
+from __future__ import annotations
+
+import re
+import sys
+from pathlib import Path
+
+# Decode boundaries: the modules where an error mistaken for "missing data"
+# becomes a silently wrong profile. Keep this list narrow and intentional.
+AUDITED_PATHS = [
+    "crates/dataprof-db/src/connectors",
+    "crates/dataprof-db/src/streaming.rs",
+    "crates/dataprof-db/src/lib.rs",
+    "crates/dataprof-parquet/src",
+    "crates/dataprof-json/src",
+    "crates/dataprof-csv/src",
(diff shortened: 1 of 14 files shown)


AndreaBozzo:
Addressed all four Copilot findings in 13f3157: invariant-backed default fallbacks now use explicit expect() messages. Validation: decode-audit guard, focused metrics/engines/python tests (183 passed), and focused clippy all pass.

PR #415, opened 2026-07-16
feat(report): version the serialized ProfileReport schema

### LINKED ISSUE #374: Version the serialized ProfileReport schema and define compatibility
Parent: #371

## User problem

`ProfileReport.save()`, `load()`, `to_dict()`, and Rust serde output make reports durable product artifacts, but the serialized document has no explicit schema version or documented reader compatibility policy. A future field rename can turn a saved baseline into an opaque load failure—or worse, a plausible but semantically different report.

## Proposed direction

Add an explicit report schema version, independent of the package version, and define how readers handle:

- the current unversioned 0.9 document
- additive fields from newer writers
- unsupported future major schema versions
- missing optional metrics (`None` means not analyzed; empty means analyzed with no findings)

Prefer a small top-level integer/string and explicit migrations over coupling the file format to `dataprof.__version__`.

## Acceptance criteria

- [ ] Newly serialized JSON reports carry a schema version.
- [ ] Current 0.9 unversioned fixtures still load through a documented legacy path.
- [ ] Unknown additive fields do not break a compatible reader.
- [ ] Unsupported future schema versions fail explicitly before partial decoding.
- [ ] Rust serde and Python `to_dict()/from_dict()/save()/load()` agree on the field and semantics.
- [ ] Golden fixtures cover legacy, current, additive-future, and incompatible-future documents.
- [ ] The compatibility policy is documented in the Python guide and Rust API docs.

## Likely files

- `crates/dataprof-core/src/profile.rs`
- `crates/dataprof-runtime/src/profile_report.rs`
- `python/dataprof/__init__.py` and `__init__.pyi`
- `tests/` and `python/tests/test_python_api.py`
- `docs/python/README.md`

## Verification

```bash
cargo test -p dataprof-core
cargo test -p dataprof-runtime
uv run pytest python/tests/test_python_api.py -q
```

## Non-goals

- guaranteeing byte-for-byte identical JSON
- embedding raw input data
- accepting semantically incompatible reports silently

AndreaBozzo:
Closes #374.

## What

Saved reports are durable artifacts (CI baselines, drift references, agent inputs), but the serialized document had no schema version: a future field change could turn a stored baseline into an opaque failure or a plausible-but-different report. This adds an explicit, package-version-independent `schema_version` to both serialization dialects and defines the reader policy.

- **Rust**: `REPORT_SCHEMA_VERSION` (= 1) exported from `dataprof-runtime` and the facade; `ProfileReport` serializes the field and its deserializer rejects newer versions with an actionable error. Documents without the field (0.9-era) load as version 0 through the existing legacy aliases.
- **Python**: `dataprof.REPORT_SCHEMA_VERSION` (= 1); `to_dict()`/`to_json()`/`save()` stamp the document; `from_dict()`/`from_json()`/`load()` check the version **before** any structural decoding — legacy documents load, additive unknown fields are ignored, newer versions raise `ValueError` ("Upgrade dataprof to load it"), non-integer versions are rejected.
- **Docs**: compatibility policy table in the Python guide; policy doc comments on the Rust const/field.

## Compatibility semantics

| Document | Reader behavior |
|---|---|
| No `schema_version` (pre-0.10) | Loads via legacy path, identifiable as version 0 |
| `schema_version` ≤ current | Loads; unknown additive fields ignored |
| `schema_version` > current | Explicit failure, never partial decoding |

## Tests

Fixture coverage in both languages for all four document classes (legacy / current / additive-future / incompatible-future), plus non-integer version rejection and a gate-ordering test (version check precedes structural validation).

## Verification

Drove the flow end-to-end through both public surfaces (not just tests): profiled a real CSV, saved, inspected the on-disk document (`schema_version: 1` leads the document), reloaded, then tampered the file on disk — version 99 rejected with the actionable message, field removed loads as legacy, string version rejected, additive unknown field tolerated. Same probes through the Rust facade via `serde_json`.

```
cargo test -p dataprof-runtime      # 41 passed
uv run pytest python/tests/test_python_api.py -q   # 265 passed
cargo fmt --all && cargo clippy -p dataprof-runtime -p dataprof -p dataprof-python --all-targets
uv run ruff format python/ && uv run ruff check python/ && uv run ty check python/
```

### DIFF
--- a/crates/dataprof-runtime/src/lib.rs
+++ b/crates/dataprof-runtime/src/lib.rs
@@ -17,7 +17,7 @@ pub use profile_builder::{
     profile_from_stats, profile_from_stats_with_hints, profiles_from_streaming,
     profiles_from_streaming_with_hints, quality_check_samples,
 };
-pub use profile_report::ProfileReport;
+pub use profile_report::{ProfileReport, REPORT_SCHEMA_VERSION};
 pub use report_assembler::ReportAssembler;
 pub use streaming_stats::{
     StreamingColumnCollection, StreamingStatistics, TextLengthStats, WelfordAccumulator,
--- a/crates/dataprof-runtime/src/profile_report.rs
+++ b/crates/dataprof-runtime/src/profile_report.rs
@@ -139,7 +139,6 @@ impl<'de> serde::Deserialize<'de> for ProfileReport {
     where
         D: serde::Deserializer<'de>,
     {
-        use serde::Deserialize;
         use serde::de::Error;
 
         let value = serde_json::Value::deserialize(deserializer)?;
--- a/crates/dataprof-runtime/src/profile_report.rs
+++ b/crates/dataprof-runtime/src/profile_report.rs
@@ -23,13 +23,12 @@ pub const REPORT_SCHEMA_VERSION: u32 = 1;
 /// Quality assessment informed by ISO 8000/25012 concepts. This is the primary output of all
 /// profiling operations (`Profiler::analyze_file`, `Profiler::analyze_source`,
 /// `Profiler::profile_stream`, etc.).
-#[derive(Debug, Clone, serde::Serialize, serde::Deserialize)]
+#[derive(Debug, Clone, serde::Serialize)]
 pub struct ProfileReport {
     /// Version of the serialized report schema (see [`REPORT_SCHEMA_VERSION`]).
     ///
     /// `0` means the document predates schema versioning (a 0.9-era report).
     /// Deserialization rejects versions newer than [`REPORT_SCHEMA_VERSION`].
-    #[serde(default, deserialize_with = "deserialize_schema_version")]
     pub schema_version: u32,
     /// Unique identifier for this report (UUID v4)
     pub id: String,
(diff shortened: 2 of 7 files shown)


AndreaBozzo:
Addressed both Copilot findings in e61b9e1 — and importantly, **the autofix commit dea26bd had broken the branch**: it deleted the structural-check `if` line but left its `raise` body dangling, so every document carrying a valid `schema_version` failed to load (13 test failures). Fixed properly:

- **null bypass (Python)**: `from_dict()` now distinguishes a missing field (legacy path) from an explicit `schema_version: null`, which is rejected as malformed. The structural key check is restored. Parametrized test now includes `None`.
- **field-order gate (Rust)**: replaced the per-field `deserialize_with` hook with a manual `Deserialize` that buffers the document into a `serde_json::Value` first (same JSON-centric approach the existing quality-compat deserializer already uses), so the actionable "upgrade dataprof" error wins even when a future document also breaks structure and `schema_version` appears last. Explicit null / non-integer / negative versions are rejected with a dedicated message. New tests cover the version-error-wins-over-structural-error scenario and the malformed-version cases.

Gates: cargo test -p dataprof-runtime (44 passed), pytest (266 passed), fmt/clippy/ruff/ty clean. Also re-drove the save→tamper→load flow end to end.

PR #421, opened 2026-07-16
fix(metrics): track duplicate rows in columnar and record-batch engines

### LINKED ISSUE #417: Cross-engine parity: columnar engine skips duplicate-row detection on files with nulls, changing the quality score
Found while dogfooding on master (583e2d8). Violates the output contract in AGENTS.md: *profile numbers must be identical regardless of which engine produced them*.

## Repro

```python
import dataprof as dp
f = "examples/test_datasets/sensor_data_outliers.csv"
dp.profile(f, engine="incremental").quality_score  # 83.81
dp.profile(f, engine="columnar").quality_score     # 76.41
```

The uniqueness facts show why: incremental reports `rows_checked: 9`, columnar reports `rows_checked: 0`. Same on `sample_data.csv` (93.75 vs 87.5 on the uniqueness dimension). `large_dataset.csv` (no nulls) is identical across engines.

## Root cause

Duplicate-row detection has two paths (`uniqueness.rs`):
1. a full-stream `RowDuplicateSummary` tracker — wired only into the incremental/async engines (`streaming_stats.rs`, `incremental.rs`);
2. a sample-based exact scan that correctly refuses to run on misaligned samples (nulls never enter the reservoir, so any column with nulls breaks row alignment) and returns "not assessed".

The columnar CSV engine has no tracker, so on any file with nulls its duplicate component silently becomes "not assessed" and the uniqueness dimension renormalizes to key-uniqueness only — a different score for the same bytes.

## Direction

Wire a row-duplicate tracker into the columnar engine (the Parquet record-batch path appears to have one — `ecommerce.parquet` reports `rows_checked: 15`), or explicitly mark the dimension's duplicate component as not-assessed in a way that cross-engine parity tests catch. Extend the parity suite with a nulls-bearing fixture: the current parity gap survives precisely because the aligned-sample case works.

## Acceptance criteria

- [ ] `engine="incremental"` and `engine="columnar"` produce identical uniqueness facts and overall score on files with nulls
- [ ] Cross-engine parity tests include at least one fixture with nulls in several columns

AndreaBozzo:
Closes #417.

## Problem

`engine="incremental"` and `engine="columnar"` produced **different quality scores for the same bytes** whenever any column contained nulls (e.g. `sensor_data_outliers.csv`: 83.81 vs 76.41). Only the incremental/async engines fed a full-stream `RowUniquenessTracker`; the columnar paths fell back to the sample-based duplicate scan, which correctly refuses to run on null-misaligned reservoirs — so the uniqueness dimension silently lost its duplicate component and renormalized differently.

## Fix

A shared `observe_batch_rows()` helper (in `record_batch_analyzer`) folds every row of a `RecordBatch` into the existing `RowUniquenessTracker` using the **same length-prefixed signature encoding as the incremental engine** (null → empty field), so duplicate counts agree across engines by construction. Wired into every previously-untracked assembler site:

- `ArrowProfiler` CSV batch loop (the `engine="columnar"` headline)
- Parquet file parser and Parquet-over-HTTP (via `RecordBatchAnalyzer`)
- Python DataFrame and Arrow-capsule bindings (same analyzer)
- the dependency-free dict/rows path (`profile_columns`)

`RowUniquenessTracker` is now exported from `dataprof-runtime`.

The last site was caught by the existing test suite mid-review: wiring the DataFrame path but not the dict path made *those two* disagree (88.94 vs 91.02) — the ad-hoc-vs-pandas parity test failed exactly as designed. All input paths now assess duplicates identically.

## Tests

- `cross_engine_consistency.rs`: nulls+duplicates fixture asserting identical uniqueness facts **and identical overall score** across incremental/columnar, plus a no-nulls regression guard
- `test_engine_parity.py`: same fixture through the Python surface, plus a DataFrame/Arrow-input duplicate-tracking test

## Verification (live, not just tests)

All five input surfaces drove the same 4-row nulls+duplicate dataset to identical facts (`rows_checked=4, duplicate_rows=1, exact`): CSV both engines, dict, parquet file, pandas, polars. Original repro files now MATCH (83.81/83.81, 87.65/87.65). Scale check: 500k rows — columnar 0.37s vs incremental 0.98s with the tracker active, both flagging the HLL spill as approximate.

Gates: `cargo test --workspace` (25 suites), `pytest python/tests/` (326 passed), fmt/clippy/ruff clean.

### DIFF
--- a/crates/dataprof-parquet/src/arrow_profiler.rs
+++ b/crates/dataprof-parquet/src/arrow_profiler.rs
@@ -1,4 +1,4 @@
-use crate::record_batch_analyzer::observe_batch_rows;
+use crate::record_batch_analyzer::BatchRowTracker;
 use arrow::array::*;
 use arrow::csv::ReaderBuilder;
 use arrow::datatypes::*;
@@ -10,8 +10,7 @@ use dataprof_csv::CsvParserConfig;
 use dataprof_metrics::CardinalityEstimator;
 use dataprof_metrics::analysis::inference::{infer_type, is_null_like_token};
 use dataprof_runtime::{
-    ColumnProfileInput, ProfileReport, ReportAssembler, RowUniquenessTracker, TextLengths,
-    build_column_profile,
+    ColumnProfileInput, ProfileReport, ReportAssembler, TextLengths, build_column_profile,
 };
 use std::fs::File;
 use std::path::Path;
@@ -142,7 +141,7 @@ impl ArrowProfiler {
         // reservoirs are misaligned (any column with nulls) would silently
         // skip the duplicate component of the uniqueness dimension — breaking
         // cross-engine score parity with the incremental engine.
-        let mut row_tracker = RowUniquenessTracker::default();
+        let mut row_tracker = BatchRowTracker::default();
 
         // The Arrow CSV reader has no row cap, so enforce it here: stop once the
         // limit is reached and slice the batch that straddles it, keeping the row
@@ -166,11 +165,7 @@ impl ArrowProfiler {
             }
 
             total_rows += batch.num_rows();
-            observe_batch_rows(&mut row_tracker, &batch).map_err(|error| {
-                DataProfilerError::ArrowError {
-                    message: error.to_string(),
-                }
-            })?;
+            row_tracker.observe_batch(&batch);
 
             // Process each column in the batch
             for (col_idx, column) in batch.columns().iter().enumerate() {
--- a/crates/dataprof-parquet/src/arrow_profiler.rs
+++ b/crates/dataprof-parquet/src/arrow_profiler.rs
@@ -1,3 +1,4 @@
+use crate::record_batch_analyzer::observe_batch_rows;
 use arrow::array::*;
 use arrow::csv::ReaderBuilder;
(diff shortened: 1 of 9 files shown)


AndreaBozzo:
Both Copilot findings addressed in f78e654:

**Public surface**: `observe_batch_rows` is no longer `pub` — callers go through a crate-private `BatchRowTracker` wrapper, so nothing new leaks into the dataprof-parquet (or re-exported dataprof-engines) API.

**Formatter failure**: agreed the whole profile must not fail over an auxiliary metric, but the suggested per-cell placeholder fallback (the `process_generic_array` pattern) would be wrong *for this metric*: index-suffixed placeholders make identical rows look distinct (masking real duplicates), and constant placeholders make distinct rows look identical (fabricating duplicates) — exactly the plausible-but-wrong number class this crate bans. Instead, `BatchRowTracker` disables itself on the first unformattable batch and `summary()` returns `None`, so the duplicate component reads "not assessed" (the pre-fix behavior, but now only for genuinely unrenderable Arrow types) while the profile succeeds. The contract is documented on the type.

Gates: clippy `-D warnings` clean, cross-engine suite 13/13, Python parity + API suites 275 passed after rebuild.

Issue #426, opened 2026-07-17
Non-UTF-8 input: wire encoding detection into the default path; current error is misleading


AndreaBozzo:
## Summary

Non-UTF-8 input (e.g. latin-1, still common in European CSV exports) fails hard, and the error is misleading on every axis:

```text
RuntimeError: Analysis failed: All engines failed: All engines failed.
Primary (Incremental): I/O error: stream did not contain valid UTF-8
Check file permissions and disk space.
Fallback (Arrow): CSV parsing failed: ... invalid utf-8 in field 1 near byte index 4
Suggestion: The CSV file 'unknown' has inconsistent column counts. This often happens with:
  � Text fields containing commas without proper quoting
  ...
```

Problems, in order:

1. `"All engines failed: All engines failed."` — duplicated prefix.
2. `"Check file permissions and disk space"` — wrong advice for an encoding error.
3. The suggestion block talks about **inconsistent column counts**, which has nothing to do with the failure, names the file as `'unknown'`, and its bullet glyphs are themselves mojibake (`�`) — an encoding bug inside the encoding error message.
4. `RobustCsvParser::detect_encoding` ([crates/dataprof-csv/src/robust_csv.rs:74](../blob/master/crates/dataprof-csv/src/robust_csv.rs#L74)) exists but the default profile path never uses it.

## Expected

Either wire encoding detection into the default path (decode latin-1/windows-1252 with a report-level note), or fail with a single, correct diagnostic: file, byte offset, detected encoding guess, and the one actionable suggestion (`re-encode to UTF-8` / future `encoding=` option).

Concrete instance of the #373 error-DX work; filing separately because the encoding fallback is a behavior decision, not just message hygiene.

Found in a 0.10.0 dogfooding session.

Issue #371, opened 2026-07-11
Epic: Product Experience & Trust


AndreaBozzo:
## Outcome

Make dataprof feel like a dependable product, not only a capable profiling engine.

A user should be able to install it, understand what the installed build can do, profile unfamiliar data, interpret the result, recover from common failures, and persist or automate decisions without learning the workspace architecture.

This epic owns cross-cutting ergonomics and user-facing contracts. Format engines, metrics, agent integration, testing, and streaming keep their own epics; work belongs here when its primary outcome is a clearer and more coherent user journey.

## Product principles

- **One obvious path:** `profile()` / `Profiler` remain the main entry points.
- **Actionable failures:** errors say what failed, preserve source context, and name the next useful action.
- **Honest capabilities:** optional features are discoverable at runtime; unsupported does not look like broken.
- **Interpretable output:** reports explain important signals without transforming or cleaning the data.
- **Stable artifacts:** saved reports have an explicit compatibility story.
- **Parity:** the same input produces the same profile semantics across engines and Python/Rust paths.
- **Local and deterministic:** no telemetry, hosted service, data movement, or implicit network calls.

## Delivered foundations

- [x] #372 — installed capability discovery
- [x] #373 — consistent actionable error contract
- [x] #374 — serialized report schema version and compatibility
- [x] #381 — duplicate normalized column-name policy
- [x] #382 — malformed JSONL policy parity
- [x] #383 — approximation provenance for distinct counts
- [x] #384 — empty quality-dimension selection contract
- [x] #386 — remove the columnar 1,000-value cardinality cap
- [x] #387 — deterministic/honest large-file distribution statistics

### 0.10 — close correctness gaps found by dogfooding

- [x] #429 — suppress locale-pattern false positives in ordinary identifiers
- [x] #459 — make sampling implement-or-reject across public paths
- [x] #460 — enforce execution-control and provenance invariants
- [x] #462 — report async ragged CSV recovery honestly
- [x] #464 — audit every published feature graph and disposition the RSA advisory

The bounded fix bundle shipped in PR #466. Final 0.10 dogfood closure is tracked by #480, #482, and #483; crate navigation is #306.

### 0.11 — the report tells the truth

0.11 shipped the scoring and provenance-honesty work found by dogfooding across sixteen input paths.

- [x] #544 — quality score is non-monotonic: 20% junk scores 100.0, 18% scores 95.5
- [x] #436 — completeness semantics for sparse optional columns
[...]

PR #446, opened 2026-07-19
feat(errors): standardize actionable diagnostics and preserve source context

### LINKED ISSUE #373: Product errors: preserve source context and standardize actionable diagnostics
Parent: #371

## User problem

`DataProfilerError` has categories and some helpful suggestions, but the contract is inconsistent:

- generic `From<std::io::Error>`, `From<anyhow::Error>`, and `From<csv::Error>` conversions replace the source path with `"unknown"`
- `UnsupportedFormat` says only CSV/JSON/JSONL are supported although Parquet is a default feature
- some messages still use legacy “DataProfiler” naming
- Python callers sometimes receive a native exception class and sometimes a generic mapped runtime error
- suggestions are embedded in display strings rather than consistently available as structured context

This makes common failures harder to recover from and makes programmatic handling fragile.

## Outcome

Define and implement a compact error contract shared by public Rust entry points and mapped deliberately at the Python boundary. Every user-facing error should preserve what operation/source failed where safe, expose a stable category, and provide a next action when one exists.

## Work

1. Inventory public error paths for CSV, JSON/JSONL, Parquet, configuration, database, and Python dispatch.
2. Write the contract before changing variants: stable category, human message, optional source/operation, optional suggestion, and error source chain.
3. Replace context-free `From` conversions at call sites where the path/operation is known.
4. Correct stale format/product wording.
5. Map predictable conditions to idiomatic Python exceptions without leaking credentials or raw data.

## Acceptance criteria

- [ ] A missing, unreadable, malformed, and unsupported-format fixture has an asserted Rust category/message and Python exception/message.
- [ ] Known paths are never rendered as `"unknown"`.
- [ ] Supported-format suggestions reflect enabled features or avoid making a false exhaustive claim.
- [ ] Database errors redact credentials and query values.
- [ ] Existing error sources remain chained where possible.
- [ ] Public error changes are documented as compatibility-sensitive.

## Likely files

- `crates/dataprof-core/src/errors.rs`
- parser and facade call sites under `crates/dataprof-*/src/`
- `crates/dataprof-python/src/`
- `tests/` and `python/tests/`
- `docs/guides/getting-started.md`

## Verification

```bash
cargo test -p dataprof-core
cargo test -p dataprof-json
cargo test -p dataprof-python
uv run pytest python/tests/test_python_api.py -q
```

## Non-goals

- automatic recovery that changes input semantics
- logging raw rows, credentials, or full SQL queries
- a new error-reporting service

AndreaBozzo:
Closes #373. Part of the 0.10.0 contract & correctness milestone (Epic #371).

Gives public error paths a consistent contract: real source paths instead of `'unknown'`, honest supported-format claims, structured suggestions, redacted credentials, and idiomatic Python exceptions.

## Changes

**`dataprof-core` — the contract**
- The context-free `From<io::Error>` / `From<anyhow::Error>` / `From<csv::Error>` conversions no longer fabricate `FileNotFound { path: "unknown" }`; they preserve the original message and let call sites attach the path. `csv_parsing()` now takes `Option<&str>` and the suggestion drops the misleading `'unknown'` filename.
- `UnsupportedFormat` advertises only formats the running build can read — Parquet only when the `parquet` feature is enabled (wired a message-only `dataprof-core/parquet` flag forwarded from the facade's `parquet`).
- New `DataProfilerError::suggestion() -> Option<String>` exposes the next action as structured data.
- New `redact_credentials()` scrubs `scheme://user:pass@host`; applied in `database_connection()` / `database_query()`.
- Fixed stale "DataProfiler" wording.

**`dataprof-csv`** — every `csv_parsing` call passes the real `file_path` (was `"unknown"` / `"current file"`).

**`dataprof-db`** — scheme-mismatch config errors report the detected scheme instead of echoing the raw connection string.

**`dataprof` facade** — `analyze_file` fails fast with a `FileNotFound` naming the real path when the file is missing, uniform across engines/formats.

**`dataprof-python`** — categories map to `FileNotFoundError`, `ValueError` (unsupported format / config / hints), `PermissionError` / `OSError`, else `RuntimeError`; the `infer_schema` / `quick_row_count` / `analyze_structure` bindings now route through the shared mapper.

## Acceptance criteria
- [x] Missing / unreadable / malformed / unsupported-format fixture → asserted Rust category/message and Python exception/message
- [x] Known paths are never rendered as `"unknown"`
- [x] Supported-format suggestions reflect enabled features
- [x] Database errors redact credentials
- [x] Public error changes documented as compatibility-sensitive (release notes)

Existing source-chain preservation was left as-is where a `From` boundary has no source to chain ("where possible"); the encoding default-path decision remains #426.

## Compatibility
Error messages and, on the Python side, exception **types** change. `except RuntimeError:` blocks that caught missing-file / unsupported-format / config errors need updating to the idiomatic classes. Documented in `docs/release-notes.md`.

## Verification
```
cargo test -p dataprof-core -p dataprof-csv -p dataprof-db --features parquet
cargo test --features parquet --test error_contract --test public_api_facade
uv run pytest python/tests/test_python_api.py -q
uv run ruff check python/ && uv run ty check python/
```
All green.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -202,6 +202,7 @@ arrow = [
 ]
 parquet = [
     "arrow",
+    "dataprof-core/parquet",
     "dataprof-parquet/parquet",
     "dataprof-engines/parquet",
     "dataprof-partial/parquet",
--- a/crates/dataprof-core/Cargo.toml
+++ b/crates/dataprof-core/Cargo.toml
@@ -32,3 +32,6 @@ default = []
 arrow = ["dep:arrow"]
 async-streaming = ["dep:tokio"]
 database = []
+# Message-only flag: lets error diagnostics advertise Parquet support accurately
+# when the workspace is built with the facade's `parquet` feature. Enables no code.
+parquet = []
--- a/crates/dataprof-core/src/errors.rs
+++ b/crates/dataprof-core/src/errors.rs
@@ -496,7 +496,7 @@ impl DataProfilerError {
                 Some("Check that the file exists and you have permission to read it.".to_string())
             }
             DataProfilerError::UnsupportedFormat { .. } => Some(format!(
-                "This build reads {}. Re-encode the input, or rebuild with the needed feature.",
+                "This build reads {}. Convert the input to one of these formats, or rebuild with the feature for the format you need.",
                 supported_formats_hint()
             )),
             DataProfilerError::MemoryLimitExceeded => {
--- a/crates/dataprof-core/src/errors.rs
+++ b/crates/dataprof-core/src/errors.rs
@@ -1,5 +1,50 @@
 use thiserror::Error;
 
+/// The file formats this build can actually read, reflecting compiled features.
+///
+/// The list must never over-claim: Parquet is only advertised when the `parquet`
+/// feature is enabled, so an "unsupported format" error cannot point the user at
+/// a format the installed build cannot open.
+fn supported_formats_hint() -> &'static str {
+    #[cfg(feature = "parquet")]
+    {
+        "CSV, JSON, JSONL, Parquet"
+    }
+    #[cfg(not(feature = "parquet"))]
+    {
+        "CSV, JSON, JSONL"
+    }
+}
+
+/// Redact credentials from a string before it enters an error message.
+///
+/// Connection strings and driver errors can carry `scheme://user:password@host`.
+/// We collapse the userinfo to `***` so passwords (and usernames) never reach a
+/// log line or a surfaced diagnostic, while keeping the scheme/host for context.
+pub fn redact_credentials(input: &str) -> String {
+    // Match the `://[userinfo@]` segment of any URL-like token and drop userinfo.
+    let mut out = String::with_capacity(input.len());
+    let mut rest = input;
+    while let Some(scheme_idx) = rest.find("://") {
+        let after_scheme = scheme_idx + 3;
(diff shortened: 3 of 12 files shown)


AndreaBozzo:
The deferred source-chain preservation (retaining `#[source]` through the `From` conversions) is tracked separately in #447 — it's an unbounded refactor gated on the `Clone` derive on `DataProfilerError`, out of scope for this contract PR.

AndreaBozzo:
Addressed all four Copilot review comments in 1b8165a:

1. **`symlink_metadata(path).is_err()` over-matched** — the fast-path now only short-circuits to `FileNotFound` on `ErrorKind::NotFound`; permission and other I/O errors fall through so the engine surfaces the accurate category (no more `PermissionError` masquerading as `FileNotFoundError`).
2. **Case-sensitive permission check** — the Python `IoError` → `PermissionError` mapping now lowercases before matching.
3. **Misleading "Re-encode" suggestion** — `UnsupportedFormat.suggestion()` now says "Convert the input to one of these formats, or rebuild with the feature for the format you need."
4. **Silently-passing malformed test** — replaced with a deterministic non-UTF-8 CSV case (the #426 scenario) that always hits the error path via `expect_err`, so it can't go vacuous if the parser becomes permissive.

fmt, clippy `-D warnings`, and the Rust + Python test suites are green.

PR #453, opened 2026-07-22
fix(partial): remove prefix bias from CSV quick_row_count sampling

### LINKED ISSUE #428: quick_row_count overestimates by 9% on autoincrement-id CSVs (prefix-biased row-size sample)
## Summary

`quick_row_count` overestimated a plain uniform CSV by **+9.3%** (`count=2,185,401` for a file with exactly 2,000,000 rows, `exact=false, method='sampling'`).

## Likely cause

The estimate extrapolates average row length from a prefix of the file. In any dataset with an autoincrement id (or growing timestamps), early rows are systematically *shorter* than later ones (`id=5` vs `id=1999999`), so file_size / prefix_avg_row_len overestimates. This is the common case, not a corner case — most real tables have a short-early-rows bias.

## Expected

Sample row lengths at several offsets across the file (seek to k evenly spaced positions, resync to the next newline, measure a few rows at each). Same I/O budget, removes the systematic bias. Also worth reporting a confidence/interval hint in `RowCountEstimate` so downstream users know the expected error.

Found in a 0.10.0 dogfooding session: `RowCountEstimate(count=2185401, exact=false, method='sampling', time=1ms)` on a 124 MB, 2,000,000-row CSV.

AndreaBozzo:
Closes #428.

## Problem

`quick_row_count` estimated a large CSV's row count from the average line length of its first 10k lines, then extrapolated `file_size / prefix_avg_row_len`. On the common case of an autoincrement id or growing timestamp, early rows are systematically shorter than later ones, so the prefix average is too small and the count is overestimated — **+9.3%** on a 124 MB / 2,000,000-row file in a 0.10.0 dogfooding session.

## Fix

Sample local **row density** (lines per byte) at 16 evenly-spaced windows across the file and integrate over the file size. Two properties make this unbiased for any row-length distribution, at the same I/O budget as the old sampler:

- **Density, not length.** Each window represents an equal share of the file's *bytes*, so averaging per-window density is a Riemann sum of rows over the byte axis. Averaging line *lengths* instead is length-biased — uniform byte offsets over-weight long rows and undercount.
- **Midpoint rule.** Windows are probed at the *center* of each byte slice, not its leading edge. Under a monotone gradient a left-edge sample is a left Riemann sum with O(1/W) bias; centered samples give the O(1/W²) midpoint rule.

## Confidence hint

Adds `RowCountEstimate.relative_error: Option<f64>` — an approximate 1-sigma relative standard error derived from the spread of per-window densities (conservative under a steep gradient; tiny for uniform files). Exposed through the Python getter, `.pyi` stub, and `repr`. `None` for exact counts. `#[serde(default)]` keeps pre-existing serialized reports deserializable.

## Scope

The JSONL sampling path shares the identical prefix bias; it is left unchanged here (with a code comment) to keep this scoped to the reported CSV case. Worth a follow-up.

## Verification

- End-to-end on a 12 MB growing-id CSV: estimate **2.0%** off (was ~9%), `relative_error=0.13`, `method='sampling'`.
- New Rust tests: uniform-file accuracy, growing-rows unbiased, exact-count has no `relative_error`. Plus a Python test that exact counts expose `relative_error is None`.
- Gates run locally: `cargo clippy --all --all-targets -- -D warnings`, `cargo fmt --all --check`, `cargo test -p dataprof-partial -p dataprof-core -p dataprof`, `uv run ruff format/check python/`, `uv run ty check python/`, `uv run pytest python/tests/test_python_api.py -k row_count`.


AndreaBozzo:
Thanks @copilot — both comments were the same real issue (a zero-sample branch returning `count: 0` labelled `exact` / `FullScan` without scanning), addressed in 27eaf08:

- **CSV (line ~487):** when `sample.lines_sampled == 0` the code now falls back to a real full scan via the new `full_scan_csv` helper (reusing the reader still positioned at offset 0), instead of returning a fabricated zero.
- **JSONL (line ~683):** same treatment via `full_scan_jsonl` — a prefix of blank lines longer than the sample no longer yields an exact zero; it full-scans the file.

Both helpers keep `exact`/`method` truthful to the work actually done. Added two regression tests: a >10 MB single-huge-line CSV (→ exact count 1) and a JSONL file with a blank-line prefix longer than the sample (→ exact count of the real records).

PR #454, opened 2026-07-22
fix(partial): apply multi-offset row-count sampling to JSONL

AndreaBozzo:
Follow-up to #453 / #428.

## Context

#453 removed the prefix bias from CSV `quick_row_count` sampling but deliberately left the JSONL sampling path on the old first-N-lines estimator, to keep that PR scoped to the reported CSV case. This closes that gap so the whole row-count sampler is consistent.

## Change

Generalize `sample_row_density_multi_offset` with a `counts_as_row: impl Fn(&str) -> bool` predicate:

- Every physical line contributes its **bytes** to the local density denominator.
- Only lines the predicate accepts count toward the **row total** — every line for CSV, non-empty lines for JSONL.

The JSONL sampling branch now uses the multi-offset density sampler (midpoint-rule windows across the whole file) and emits the same `relative_error` hint the CSV path does.

Two nice consequences fall out of sampling the whole file instead of the prefix:
- A run of blank lines no longer skews the estimate — the blanks charge bytes without adding rows.
- The old "prefix is all blanks → fabricated exact zero" case disappears; the exact-scan fallback now only triggers when *no window* observes a record (e.g. an essentially all-blank file), and still returns a truthful full-scan count.

## Verification

- End-to-end on a 12 MB growing-record JSONL: estimate **0.66%** off (`relative_error≈0.10`, `method='sampling'`), versus the ~9%-class prefix bias before.
- New Rust tests: JSONL growing-rows accuracy, a blank-prefix file estimated correctly, and an all-blank file falling back to an exact scan.
- Gates run locally: `cargo clippy --all --all-targets -- -D warnings`, `cargo fmt --all --check`, `cargo test -p dataprof-partial`, `uv run pytest python/tests/test_python_api.py -k "row_count or count_json or jsonl"`.

### DIFF
--- a/crates/dataprof-partial/src/lib.rs
+++ b/crates/dataprof-partial/src/lib.rs
@@ -454,8 +454,8 @@ fn count_csv(path: &Path, start: Instant) -> Result<RowCountEstimate, DataProfil
         // case of an autoincrement id or growing timestamp, where early rows
         // are shorter than later ones, so extrapolating from the file's head
         // overestimates. Sampling across offsets removes that bias for the same
-        // I/O budget.
-        let sample = sample_row_density_multi_offset(path, file_size)?;
+        // I/O budget. Every physical line counts as a row for CSV.
+        let sample = sample_row_density_multi_offset(path, file_size, |_| true)?;
 
         if sample.lines_sampled == 0 {
             // The sampler could not observe a single complete line — e.g. rows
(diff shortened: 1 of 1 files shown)


AndreaBozzo:
Thanks @copilot — both comments flag the same thing: the sample fields count *rows* now (JSONL skips blanks), so the `*_lines` names were misleading. Renamed in 4b350aa:

- `MultiOffsetLineSample` → `MultiOffsetRowSample`
- `estimated_lines` → `estimated_rows`, `lines_sampled` → `rows_sampled`
- the `total_lines` accumulator → `total_rows`

Callers updated; `malformed_lines` (parser line diagnostics) left as-is since it genuinely counts lines. No behavior change; 29 partial tests still pass.

PR #455, opened 2026-07-22
feat(python): finish #435 report API ergonomics — profiles, save html/md, low_sample_warning, zero-row semantics

### LINKED ISSUE #435: Python report API ergonomics: column_profiles iteration, save() hints, zero-row semantics
## Summary

Small Python report-API ergonomics issues from the 0.10.0 dogfooding session. Batched here because each is a one-liner to discuss but they share a theme: the report surface should never surprise a notebook user. Any that are breaking should be decided **before** 0.10.0 ships; the rest can trail.

## 1. Iterating `column_profiles` yields names, not profiles (possibly breaking)

```python
for col in report.column_profiles:
    col.name   # AttributeError: 'str' object has no attribute 'name'
```

Dict-like iteration yields keys. Fine for `report.column_profiles["amount"]`, but the natural loop reads like a list of profiles. Either make iteration yield profiles (breaking), or add `report.columns_iter()` / make `.columns` the list-of-profiles accessor and document the mapping semantics loudly.

## 2. `save("report.html")` error omits the HTML answer

`ValueError: Unsupported format. Use .json, .csv, or .parquet` — but `to_html()` exists. The error should say so (and arguably `save()` should just support `.html`/`.md` since the render methods exist).

## 3. Zero-row files report "analyzed" values instead of absent ones

Header-only CSV: every column comes back `string` with `null_percentage: 0.0`, `avg_length: 0.0`, `unique: 0`. Per the None-vs-empty contract ("None/absent = not analyzed"), stats computed over zero values should be absent, not zero — `null% 0.0` on a column with no rows reads as "no nulls", which was never measured.

## 4. `DataQualityMetrics.low_sample_warning` round-trip

`low_sample_warning` is exposed as a property and read by `from_dict()`, but `to_dict()` output doesn't visibly include it at quality top-level — verify it serializes, and add it to the schema doc if kept.

From the 0.10.0 dogfooding session (2026-07-17).

AndreaBozzo:
**Closes #435** — all four items of the 0.10.0 dogfooding ergonomics batch.

## 1. `column_profiles` iteration (non-breaking)

`for col in report.column_profiles: col.name` raised `AttributeError` because `column_profiles` is a `name → ColumnProfile` mapping (iterating it yields names). Existing tests rely on that mapping contract, so rather than break it, this adds:

```python
for col in report.profiles:      # ordered list[ColumnProfile]
    print(col.name, col.null_percentage)
```

`column_profiles`' docstring now states the mapping semantics and points to `.profiles`.

## 2. `save()` HTML / Markdown

`save()` now writes `.html` (`to_html()`) and `.md`/`.markdown` (`to_markdown()`); the unsupported-extension error names the path and every supported format.

## 4. `low_sample_warning` round-trip

`to_dict()` now **always** emits `quality.low_sample_warning` (True *and* False) — a non-optional bool should never require inferring absence. Round-trips through `from_dict()`; documented in the report-schema notes. Additive, no schema bump.

## 3. Zero-row semantics (the one I'd initially deferred — now done)

A header-only / zero-row column has no values to measure, so ratios and aggregates were reported as `0.0`, reading as "0% nulls" / "avg length 0" when nothing was measured. Fixed at the `PyColumnProfile` boundary (shared by every engine, so output parity holds):

- `null_percentage` and `uniqueness_ratio` are now `Option<f64>` → `None` when `total_count == 0` (the ratio is undefined, not zero).
- text `min_length`/`max_length`/`avg_length` are absent for a zero-row column.
- raw counts (`total_count`, `null_count`, `unique_count`) stay `0` — genuinely "analyzed, found none". (Principled split: division-by-zero derived stats are absent; counts are a real zero.)

Kept consistent everywhere:
- The standalone Arrow export (`analyze_csv_to_arrow`/`analyze_parquet_to_arrow`) marks those two fields nullable and emits `null` for zero-row columns, so it agrees with the report API instead of diverging to `0.0`.
- `repr`/`to_html`/`to_markdown` render an em dash for an absent percentage (new `_pct_str` helper) instead of crashing on `None`.
- Stub types and the `messy_csv` example updated for the now-optional fields.

## Verification

- New tests: `.profiles` accessor; `save()` html/md; `low_sample_warning` both states; a `TestZeroRowSemantics` class covering absent-vs-zero across attribute access, `to_dict()`, Arrow export, `from_dict` round-trip, and crash-free rendering. Updated the two existing tests that encoded the old "omit False" / "`.html` unsupported" behavior.
- Full green: `pytest python/tests/test_python_api.py` (296 passed), `ruff format --check`, `ruff check`, `ty check` (caught and fixed an example that assumed `float`), `cargo clippy -p dataprof-python --all-targets -D warnings`, `cargo fmt --check`. The `messy_csv_inspection.py` example runs clean.


AndreaBozzo:
Thanks @copilot — both addressed in dba848f:

1. **Arrow `uniqueness_ratio` scale** (good catch, real cross-surface bug): the export scaled it `* 100` while `ColumnProfile.uniqueness_ratio` and the docs define it as `unique_count / total_count` in `[0, 1]`. Removed the `* 100` so `interop.analyze_csv_to_arrow()` agrees with the report API. Added a regression test asserting the exported ratio equals `unique_count / total_count` from the same batch and stays within `[0, 1]` (it would have failed at the old 0–100 scale). No test depended on the old scale.
2. **`save()` error hint**: now lists `.markdown` alongside `.md`, matching what the method accepts.

Full suite green (297), plus ruff/ty/clippy/fmt.

Issue #509, opened 2026-08-01
Decide and document the end state of the deprecated flat DataQualityMetrics accessors


AndreaBozzo:
Parent: #371

## Problem

The flat `DataQualityMetrics` accessors were deprecated in 0.9.0 (#326, implemented
in #350). The changelog entry says:

> The flat accessors keep their 0.8-compatible values **for the 0.9 line**.

0.9 and 0.10 have both shipped and 0.11 is in progress, but no removal decision was
ever recorded. Sixteen accessors are still live, still warning, and still the
shortest path to the evidence behind a dimension score:

`complete_records_ratio`, `data_type_consistency`, `duplicate_rows`,
`encoding_issues`, `format_violations`, `future_dates_count`,
`high_cardinality_warning`, `invalid_date_values`, `key_uniqueness`,
`missing_values_ratio`, `negative_values_in_positive`, `null_columns`,
`outlier_ratio`, `range_violations`, `stale_data_ratio`, `temporal_violations`.

The warning is emitted from two places that must stay in agreement:
`crates/dataprof-python/src/types.rs` for the native `DataQualityMetrics`, and
`_DictQuality._warn_flat_accessor` in `python/dataprof/__init__.py` for reports
reconstructed from `to_dict()`.

## Why it surfaced now

Writing the agent skill reference, the evidence fields were documented straight
off `dir(DataQualityMetrics)` — which lists the deprecated accessors alongside the
current ones with nothing to distinguish them. The reference shipped teaching a
deprecated path, and only a `DeprecationWarning` in an unrelated test run caught
it. Anything that discovers this API by introspection has the same problem:

- `dir()` does not mark them.
- The stubs do not mark them.
- Nothing in the repo pins how long they live.

A deprecation with no recorded end date is indistinguishable from a supported API
that happens to be noisy.

## Outcome

The deprecation has a decided, documented, and enforced end state.

## Work

- Decide the terminal release for the flat accessors and record it in
  `docs/release-notes.md`. If they are staying indefinitely, say that instead —
  an undated deprecation should be either scheduled or withdrawn.
- Make the deprecation discoverable without triggering it: mark the accessors in
  `python/dataprof/__init__.pyi` (`@deprecated`, PEP 702) so type checkers and
  editors flag them, and name the replacement key in each docstring.
- Keep the native and `_DictQuality` deprecation sets identical. They are
  maintained separately today and can silently diverge; add a test asserting the
  two sets match.
[...]

Issue #362, opened 2026-07-09
Epic: Agent & LLM Integration


AndreaBozzo:
Tracking issue for the Agent & LLM Integration epic. This replaces the milestone
of the same name: GitHub allows only one milestone per issue, so themes now live
in epics + the `area: agents` label, and milestones are reserved for releases.

## Positioning (preserved verbatim from the milestone)

> NOTE: #331 (to_llm_context) and #332 (redaction) have been pulled forward into the 0.9.0 milestone — they're useful standalone (paste-into-chat) and privacy-critical. They remain prerequisites for the MCP work tracked here. Positioned AFTER 0.9.0 (target ~0.10.0): the MCP tool depends on analyze_structure() (#319) and a stable compare() from 0.9.0. Positioning: dataprof is the local, deterministic data-profiling layer your AI coding agent (Claude Code, Codex, Cursor) can safely call before touching a dataset — it produces bounded, safe, useful profiling CONTEXT that agents consume, NOT an 'AI data quality assistant'. Build order: [in 0.9.0] #332 redaction, then #331 to_llm_context (core output contract). [here, 0.10] #336 security checklist (gates MCP); #330 MCP server (build LAST — needs #331/#332/#336/#319); then #335 column_context, #334 docs; #333 knowledge snippets (small). Non-goal: reintroducing a CLI (retired; Python + MCP is more coherent).

## Status

The two prerequisites called out above **shipped in 0.9.0**:

- #332 redaction — `to_llm_context()` now fails closed when it cannot prove a
  column is free of sensitive patterns.
- #331 `to_llm_context(max_tokens=...)` — the core output contract.

`analyze_structure()` (#319) and `compare()` also shipped, so the original reporting
prerequisites are available; execution bounds in #439 still gate the MCP server.

## Children

- [x] #336 Agent/MCP security hardening checklist — shipped in 0.10 via #410
- [ ] #330 dataprof MCP server — build last
- [ ] #334 Docs: "Using dataprof with AI agents"
- [ ] #439 Bound agent profiling cancellation and concurrency — gates #330
- [ ] #335 Progressive disclosure: per-column drill-down

## Timing: stable SDK gate satisfied; execution bounds still required

Re-verified on 2026-09-05 against the official SDK:

- Stable [v2.1.1](https://github.com/modelcontextprotocol/python-sdk/releases/tag/v2.1.1) is released.
[...]

Issue #330, opened 2026-07-05
dataprof MCP server (python -m dataprof.mcp, pip install dataprof[mcp])


AndreaBozzo:
Parent: #362. Not milestoned: moved out of 0.12.0 to the P2 backlog on 2026-09-10 so findings (#375), the quality gate (#376) and its pipeline entrypoint (#710) carry that release. See #362.

## Outcome

Ship a small, local, read-only MCP server that lets Claude Code, Codex, Cursor, and other MCP hosts profile permitted datasets through dataprof's existing deterministic and hardened surfaces.

Packaging: `pip install dataprof[mcp]`. Entry point: `python -m dataprof.mcp`. Initial transport: stdio only.

## Timing and SDK contract

Build against the official MCP Python SDK v2 after its stable release, using `MCPServer` and a dependency range bounded to the v2 major line (`mcp>=2,<3`). Do not ship against a prerelease or add an unbounded `mcp` dependency.

## Tool surface

Keep the initial tool set small and non-overlapping:

- `analyze_structure(path, max_rows=...)` — bounded first look
- `profile(path, ...)` — bounded profile and concise summary; returns an opaque report handle
- `describe_column(report_handle, name)` — detail from the already-created report, without re-profiling
- `compare(left, right, ...)` — bounded comparison of two permitted sources

Do not add a separate `quality_summary` tool unless it exposes behavior that `profile` cannot provide; overlapping tools make host selection less predictable.

## Output contract

- Return typed `structuredContent` validated by explicit JSON Schema 2020-12 output schemas, plus a concise text fallback for hosts that primarily consume text.
- Reuse the versioned report contract and #413 where the full report shape is exposed; use deliberately smaller typed models for structure, summary, comparison, and column detail.
- Keep output bounded and deterministic. No raw row dumps. Samples remain disabled unless the server policy explicitly opts in.
- Declare tools in deterministic order.
- Mark every tool accurately with MCP annotations. These tools should be read-only, non-destructive, idempotent for identical inputs, and closed-world when confined to configured local roots.
- Return recoverable tool-input failures as structured tool errors rather than leaking tracebacks or generic protocol failures.

## Report-handle lifecycle

`profile` and `describe_column` must agree on an explicit lifecycle:

- opaque, unguessable handle; never encode a host path
- bounded in-memory store with maximum entries/bytes and deterministic eviction
- bounded lifetime and explicit expired/unknown-handle errors
- no persistence to disk
[...]

PR #510, opened 2026-08-01
feat(agents): make the dataprof skill accurate, self-checking, and able to grow

AndreaBozzo:
## 📝 Summary

The Claude skill and Cursor rule had drifted from the shipped API, and nothing
detected it. Both omitted `to_llm_context()` — the only export that enforces
redaction — while the README and guides have documented it since 0.9. They also
omitted `capabilities()`, the execution trust signals, and `AgentGuard`, which
until now was documented nowhere outside its own module docstring.

This fixes the drift and adds the check that makes it fail the build next time,
then builds the skill out so it has somewhere to grow: gated reference files, a
bundled script, and evaluations.

None of this depends on the MCP server (#330) or the stable MCP SDK. It is the
part of the agent surface that can move while that work is blocked.

## 🔧 Type of Change

- [x] 🐛 Bug fix (non-breaking change that fixes an issue)
- [x] ✨ New feature (non-breaking change that adds functionality)
- [ ] 🚨 Breaking change (would cause existing functionality to not work as expected)
- [x] 📚 Documentation update
- [ ] ⚡ Performance improvement
- [ ] 🎨 Code refactoring (no functional changes)

## 📋 Changes Made

**`docs(agents)`: sync agent instructions with the shipped API**

- `SKILL.md` and `.cursor/rules/dataprof.mdc` now cover `to_llm_context()`,
  `capabilities()`, the seven trust signals (`sampling_applied`,
  `truncation_reason`, `source_exhausted`, `low_sample_warning`, `error_count`,
  `ragged_row_count`, `sampling_ratio`), and `AgentGuard`/`SandboxPolicy`.
- New `python/tests/test_agent_docs_sync.py`: every API an agent doc names must
  resolve on the installed package, every keyword in a snippet must be one the
  callee accepts, every field must exist on the type its section claims, and
  every quality-dimension key must be one dataprof actually produces.

**`docs(agents)`: split the skill into gated reference files**

- `reference/api.md` (full surface) and `reference/interpretation.md` (what each
  signal does and does not mean), linked one level deep. `SKILL.md` stays at 169
  lines; the references cost nothing until read.

**`test(agents)`: add skill evaluations**

- Three scenarios in `evals/` covering the steps agents skip most: reporting a
  ragged read honestly, keeping sensitive values local, reaching for `compare()`.
- Their rubrics quote concrete numbers, so those numbers are pinned to live
  profiling output. A rubric citing stale values grades against fiction.

**`feat(agents)`: bundle a bounded summary script**

- `scripts/dp_context.py`: summary, single column, structure-only, comparison,
  and sandboxed profiling of an untrusted path. Surfaces the trust signals and
  has no path to emitting raw cell values.
- **Not a CLI**: skill-scoped, not packaged, not installed, not a console
  script. The release surface is still crates plus wheels. Stated in the module
  docstring so it does not get promoted by accident.

**Related Issue:** part of #362; found and filed #509 along the way.

### How did you test this?

- [x] Added new tests for this feature
- [x] All existing tests pass
- [x] Tested with sample data files
- [x] Tested edge cases (traversal, missing file, conflicting flags, unassessed dimensions)

Every new assertion was proven to fail before being trusted, by injecting the
defect it guards and confirming the failure, then restoring:

| Guard | Injected defect | Caught |
| --- | --- | --- |
| attributes exist | `report.to_nonexistent_thing()` in SKILL.md | ✅ |
| keywords accepted | `dp.profile(bogus_kwarg=1)` | ✅ |
| fields exist | `bogus_field_xyz` under ColumnProfile | ✅ |
| sections reachable | renamed `## Capabilities fields` heading | ✅ |
| dimension keys | `validity['bogus_key_xyz']` | ✅ |
| no deprecated accessors | `quality.missing_values_ratio` in prose | ✅ |
| no value leaks | script echoing a raw CSV line | ✅ |
| null-ratio scaling | reverting the ×100 fix | ✅ |

The documented workflow was also executed end to end against real files:
redaction holds (including with `include_samples=True`), the guard rejects
traversal, `save`/`load` round-trips.

### Test Commands Run

```bash
uv run pytest python/tests/ -q                     # 667 passed, 5 skipped
uv run ruff format --check python/ .github/scripts/ .claude/skills/dataprof/scripts/
uv run ty check python/
```

Rust crates are untouched, so the workspace Rust gates were not run.

One run showed a `test_save_parquet` failure that did not reproduce across four
subsequent full runs, in isolation, or on stashed state — a pre-existing Windows
flake, unrelated to this branch.

## 📖 Documentation

- [x] Updated docs/ folder if architectural changes
- [x] Added/updated inline code comments

`docs/guides/agent-workflows.md` documents the new skill layout, states that the
script is not a CLI, and points at the sync test so new agent docs get added to
`AGENT_DOCS`.

## ⚡ Performance Impact

- [x] No performance impact

## 🚨 Breaking Changes

- [x] No breaking changes

## 📌 Additional Notes

**Found while writing this: #509.** The 16 flat `DataQualityMetrics` accessors
were deprecated in 0.9.0 (#326/#350) with the changelog saying they hold "for the
0.9 line", but no removal decision was ever recorded and 0.10 has shipped. They
are unmarked in `dir()` and in the stubs, so the reference here was written off
introspection and documented the deprecated path without noticing — caught only
by a `DeprecationWarning` in an unrelated test run. Both reference files now use
the nested dimension dicts, and a guard blocks agent docs from naming a
deprecated accessor.

**Not addressed here:** a sandboxed run still prints the absolute host path in
its summary, because `to_llm_context()` uses `report.source`. `AgentGuard`'s
contract covers error messages, so this is not a violation — but #330 will want
summaries scoped to the logical root too. Happy to file it separately.

**Review note:** the four commits are separable. 1 and 2 are the coherent unit
that fixes the live problem; 3, 4, and 5 build on top. Say the word if you would
rather land them as separate PRs.


AndreaBozzo:
All four review comments were correct and are fixed in dbec4ae.

**1 & 2 — error sanitization in `dp_context.py`.** Both confirmed by reproduction, and they were worse together than either alone. `--root /nonexistent/path` printed a full traceback ending in:

```
ValueError: sandbox root is not an existing directory: C:\Program Files\Git\nonexistent\path
```

— a host path, from a script whose module docstring promises it never prints raw values. The generic handler had the same shape: profiling a directory produced `Analysis failed: All engines failed: Primary (Incremental): I/O error: ...`, engine internals that can carry a path or a fragment of the row that failed to parse.

Both now route through a `_sanitize()` helper that delegates to `AgentGuard.sanitize_error()` when a guard is in play — so the two surfaces cannot drift — and applies the identical policy when it is not (no `--root`), since the script is agent-facing in both modes. Guard rejections still pass through intact:

```
$ dp_context.py x.csv --root /nonexistent/path
ValueError: request failed (details withheld at the agent boundary)

$ dp_context.py ../../../../etc/passwd --root evals/fixtures
PathNotAllowedError: no such file inside the sandbox: passwd
```

Three regression tests added, each reverted independently to confirm it fails against its own defect rather than being covered by a neighbour:

| Test | Defect reverted | Fails |
| --- | --- | --- |
| `test_invalid_sandbox_root_is_sanitized` | guard construction moved back outside `try` | ✅ |
| `test_engine_failures_do_not_echo_their_message` | handler echoes `{exc}` again | ✅ |
| `test_guard_rejections_keep_their_safe_message` | `_sanitize` flattens guard errors too | ✅ |

The third exists because the first two could otherwise be satisfied by withholding everything, which would leave an agent unable to tell why a path was refused.

**3 & 4 — Makefile divergence from CI.** Also correct, and it predated this PR: `lint-python` and `fmt-python` never covered `.github/scripts/`. I had made it worse by adding the skill path to one list and not the other. Rather than patch three call sites that can drift again, both targets now read a single list declared at the top of the Makefile:

```make
# Every path CI runs ruff over. Keep this list and the ruff steps in
# .github/workflows/ci.yml identical, or `make check` passes what CI fails.
RUFF_PATHS := python/ .github/scripts/ .claude/skills/dataprof/scripts/
```

Verified identical to the two ruff steps in `ci.yml`.

**Gates:** `ruff format --check`, `ruff check`, `ty check` all clean; 670 passed, 5 skipped.

Issue #526, opened 2026-08-03
to_llm_context flags threshold on unrounded values, so a round-trip changes the flags


AndreaBozzo:
Parent: #511 — found while writing the round-trip parity guard (#512).

## Problem

`to_llm_context()` derives its quality flags in `_column_flags()`
(`python/dataprof/__init__.py`), which compares the null-heavy threshold
against the **unrounded** `null_percentage` but renders the **rounded** one:

```python
null_pct = col.null_percentage or 0.0
if null_pct >= 100.0:
    flags.append((100.0, f"{name}: all-null"))
elif null_pct >= _NULL_HEAVY_PCT:                       # unrounded
    flags.append((null_pct, f"{name}: null-heavy ({_r2(null_pct):.1f}% null)"))  # rounded
```

Two consequences, from one cause.

**1. The flag disagrees with the number printed next to it.** A column at
19.998% null is rendered as "20.0% null" but is not flagged null-heavy,
because the threshold saw 19.998. The output states the threshold value and
withholds the threshold verdict.

**2. Saving and reloading a report changes what it reports.** The report
document stores `null_percentage` at 2dp, so the reloaded report thresholds on
20.0 and *does* flag the column. Same data, same report, different
agent-facing output:

```
--- native
+++ round-tripped
 flags (3):                          flags (4):
 - email: null-heavy (33.3% null)    - email: null-heavy (33.3% null)
                                   + - notes: null-heavy (20.0% null)
 - amount: 1 outliers (0.0%)         - amount: 1 outliers (0.0%)
```

Reproduced on a 10,501-row CSV where `notes` is 19.998% null. This is the only
`ProfileReport` member that survived the #512 parity sweep with a divergence;
every other accessor agrees across both backings.

## Why it was not fixed in #512

Fixing it changes which flags agent-facing output emits, which #512 lists as a
non-goal ("changing any reported value"). It is pinned there instead by
`test_llm_context_flags_survive_roundtrip`, an `xfail(strict=True)` that will
fail — demanding its own removal — the moment this is fixed.

## Work

- Threshold on the same value the flag renders, so the verdict and the number
  agree. The natural choice is the serialized precision, since that is what
  every consumer of the report sees.
- Keep `all-null` exact. It is a structural claim, not a threshold: rounding
  99.996% up to 100.0 would report "all-null" for a column that has values.
  Deriving it from `null_count == total_count` makes it exact and immune to
  rounding by construction.
[...]

Issue #513, opened 2026-08-01
Rounding parity: Rust and Python round the same fields differently


AndreaBozzo:
Parent: #511

## Problem

Rust and Python round the same fields with different algorithms.

- Rust, `crates/dataprof-core/src/serde_helpers.rs`: `(v * 10^n).round() / 10^n`
  — a binary float multiply, `f64::round()`, then a divide. 30 serialized
  fields use it.
- Python, `python/dataprof/__init__.py`: `Decimal(str(v))` quantized under
  `ROUND_HALF_UP`. 68 call sites use `_r2`/`_r4`.

The Python helper's docstring says it matches the Rust convention. It matches
the *intent* (half away from zero) but not the *implementation*, because
`Decimal(str(v))` reasons about the shortest decimal repr of the float while the
Rust path reasons about the binary value. They disagree whenever that repr is an
exact tie at the target precision:

```
23/4000 * 100 = 0.575  ->  Python 0.58,  Rust algorithm 0.57
```

The 0.58 is from the real Python code path. Over 200,000 random floats the two
agree 100% of the time, so this is narrow — but count-derived percentages are
exactly the shape that produces clean decimal ties, and "confidently wrong
number" is the failure class this project cares most about.

**Scope honesty:** no end-to-end case was found where the Rust and Python
serializers emit different values for the same field. The mechanism is
demonstrated; an instance is not. This issue is about closing a latent
inconsistency and proving which behaviour is intended, not about a known bug.

## Outcome

One rounding convention, with a test that fails if the two layers drift apart.

## Work

- Decide which behaviour is correct. `Decimal(str(v))` is arguably the more
  defensible for a *reported* number; `(v*100).round()/100` is what the
  serialized schema has always emitted. Whichever wins, the other changes.
- Add a property test over count-derived values (`k/n * 100` for a range of
  `n`) asserting both layers produce the same rounded result.
- Document the convention where a reader will find it, not only in a helper
  docstring.
- If the chosen convention changes any currently emitted value, say so in the
  release notes — a profiler changing a reported digit is user-visible.

## Acceptance criteria

- [ ] Rust and Python produce identical rounded output for the same input.
- [ ] A property test covers count-derived percentages, not just random floats,
      since random floats never hit the disagreement.
- [ ] The test fails if either implementation is changed in isolation.
- [ ] Any change to emitted values is documented.

## Non-goals

- Changing how statistics are computed; only how they are rounded for output.

Issue #511, opened 2026-08-01
Epic: One contract, one implementation


AndreaBozzo:
## Outcome

Every contract dataprof exposes has one implementation, or an enforced proof
that its two implementations agree. Today several are maintained by hand in two
places, with nothing asserting they match.

This epic is scoped by an audit of the workspace, not by file size. File size is
a symptom; the disease is duplicated contracts.

## What the audit found

**Four duplicated contracts.**

1. **Shadow report types.** `_DictPattern`, `_DictColumn`, `_DictQuality`, and
   `_DictBackedReport` in `python/dataprof/__init__.py` are 278 lines
   re-implementing the native report surface for reports reconstructed from
   `to_dict()`, `load()`, or `from_json()`. The deprecation list alone is
   maintained twice — in `crates/dataprof-python/src/types.rs` and in
   `_DictQuality._warn_flat_accessor` — which is #509. Nothing asserts the two
   backings agree on any value.

2. **Two rounding algorithms for the same fields.** Rust rounds 30 serialized
   fields with `(v * 10^n).round() / 10^n` (`dataprof-core/src/serde_helpers.rs`).
   Python rounds 68 sites with `Decimal(str(v))` under `ROUND_HALF_UP`
   (`_half_up`/`_r2`/`_r4`). The Python comment says it matches Rust; it matches
   the intent, not the implementation. They diverge on exact decimal ties:

   ```
   23/4000 * 100 = 0.575  ->  Python 0.58,  Rust algorithm 0.57
   ```

   The 0.58 is from the real Python path. No end-to-end instance was found where
   both serializers emit different values for one field, so this is a
   demonstrated mechanism without a demonstrated instance — a latent
   inconsistency, not a known bug. It is untested on both sides.

3. **Two producers for one documented schema.** `docs/schema/profile-report.v1.schema.json`
   can be written by Rust serde or by Python `to_dict()`. #374 versioned the
   schema; nothing asserts the producers agree.

4. **Two surfaces for one number.** Raw `ColumnProfile` attributes drift across input paths in the last two or three significant digits (three distinct values for `mean` and `std_dev` over sixteen paths), while the serialized report rounds to four decimals and agrees exactly. `AGENTS.md` says the numbers must be identical without saying which surface it governs. That is #547, and it decides whether the #526 class of bug (thresholding on unrounded values) is possible at all. Its resolution should match whichever rounding implementation this epic settles on.

**Python is missing a guard the Rust side already has.**

[...]

PR #527, opened 2026-08-03
fix(python): restore hint bindings and ratio precision on report round-trip

### LINKED ISSUE #512: Report round-trip parity: native and dict-backed reports must agree
Parent: #511

## Problem

A `ProfileReport` has two backings. One wraps the native report from the
extension; the other (`_DictBackedReport`, `_DictColumn`, `_DictQuality`,
`_DictPattern` in `python/dataprof/__init__.py`) reconstructs the same surface
from a plain dict, and is what users get from `ProfileReport.load()`,
`from_json()`, and `from_dict()`.

Nothing asserts the two agree. A user who saves a report and loads it back can
silently get different values from the same accessor, and the only reason we
know the two drift is #509 — the deprecation set is maintained separately in
each and nobody noticed until a `DeprecationWarning` surfaced in an unrelated
test run.

## Outcome

Saving and reloading a report cannot change what it reports.

## Work

- Profile a fixture, round-trip it through `to_dict()`/`from_dict()` and
  `save()`/`load()`, and assert every public member of `ProfileReport`,
  `ColumnProfile`, and `DataQualityMetrics` agrees between backings.
- Drive the member list from introspection rather than a hand-written list, so a
  new accessor is covered the day it is added rather than the day someone
  remembers to add it.
- Cover the nested quality dimension dicts key by key, not just the scores.
- Preserve the absence rule: `None` on one side must be `None` on the other, and
  must not compare equal to `0`.
- Use fixtures that exercise at least one unassessed dimension, one approximate
  count, and one column with a detected pattern.

## Acceptance criteria

- [ ] Every public accessor on the three types is compared across both backings.
- [ ] New accessors are covered without editing the test.
- [ ] Divergences found while writing this are fixed or filed, not silenced.
- [ ] The test fails when an accessor is made to disagree on purpose.

## Non-goals

- Collapsing the two backings; that is the structural child of #511 and this
  test is its safety net.
- Changing any reported value.

AndreaBozzo:
Closes #512. First of the three guards in #511.

## What this does

Adds the round-trip parity guard between the two `ProfileReport` backings, and
fixes the two defects it found on its first run.

The member lists are driven by introspection, not hand-written, so a new
accessor is covered the day it is added. `ColumnProfile` and
`DataQualityMetrics` use different classes per backing, so the sweep compares
the *union* of both sides' public members — an accessor added to one and not
the other fails rather than going unnoticed.

## The two defects

**`semantic_hint_bindings` was dropped on reload.** `to_dict()` writes it,
`save()` persists it, `_DictBackedReport` never read it. A reloaded report
reported no bindings — indistinguishable from a run profiled without hints.
One line to fix; the guard fails without it.

**`uniqueness_ratio` lost two orders of magnitude of precision.** It is a
`0..1` ratio but was rounded with `_r2`, the 2dp helper meant for `0..100`
percentages. Every column below 0.5% uniqueness serialized as a flat `0.0`:

| column | unique / total | before | after |
| --- | --- | --- | --- |
| `active` | 2 / 10,501 | `0.0` | `0.0002` |
| `notes` | 12 / 10,501 | `0.0` | `0.0011` |
| `when` | 10 / 10,501 | `0.0` | `0.001` |

Now `_r4`, matching the sibling ratio `true_ratio`, which already rounds to 4dp
— so a ratio carries the same resolution as the equivalent percentage. Both
serialization sites are fixed: `column_to_dict()` (the report document) and
`_column_record()` (dataframe/arrow/`describe`/`save`).

⚠️ **This changes emitted values** in `to_dict()`, `to_json()`, `save()` and the
dataframe exports. `docs/python/README.md` is updated; the old text stated "2
decimal places for percentages and ratios", which is what produced the bug.

## What is deliberately not fixed here

`to_llm_context()` thresholds its flags on the unrounded `null_percentage` but
renders the rounded one, so a column at 19.998% null renders "20.0% null"
without the null-heavy flag — until a round-trip stores 20.0 and it gets one.
Fixing that changes which flags agent-facing output emits, a #512 non-goal, so
it is filed as **#526** and pinned here by an `xfail(strict=True)` that turns
into a failure the moment it is fixed. It is the only accessor that diverges.

## Notes for review

- **Rounding tolerance.** Floats are not compared bit-for-bit: the dict backing
  carries values the serializer already rounded. A round-tripped float must
  equal the native value rounded at one of the documented precisions. Both
  candidate *algorithms* are accepted, so this guard does not pre-empt #513.
- **The guard does not discriminate the `uniqueness_ratio` fix**, and cannot:
  `0.0` *is* the faithful 2dp rounding, so both backings agreed. Parity held;
  the precision choice was the defect. Two dedicated tests cover it instead.
- `score_weights` is duplicated between `types.rs` and `_DictQuality`; the
  values agree today and the sweep now pins that they keep agreeing. Python
  cannot configure weights, so the duplication is latent rather than reachable.

## Verification

Each fix was reverted separately and the suite re-run — reverting both together
would only prove the tests catch *something*:

| Reverted | Fails |
| --- | --- |
| `semantic_hint_bindings` only | `test_semantic_hint_bindings_survive_roundtrip` |
| `uniqueness_ratio` 4dp only | both `test_small_uniqueness_ratio_*` tests |
| sabotage `_DictColumn.null_count` by +1 | 10 sweep tests |
| sabotage absence rule (`None` → `0`) | sweep, with `absence lost` |

Note the generic sweep does not catch the hint-bindings drop on its own — the
rich fixture supplies no hints, so both sides are `[]` there. The dedicated
hinted-report test is what discriminates it.

```
uv run pytest python/tests/ -q          # 688 passed, 5 skipped, 1 xfailed
uv run ruff check  python/ .github/scripts/ .claude/skills/dataprof/scripts/
uv run ty check python/
```

No Rust changed, so the cargo gates were not run.

### DIFF
--- a/docs/python/README.md
+++ b/docs/python/README.md
@@ -205,10 +205,17 @@ reloaded = dp.ProfileReport.from_json(report.to_json())  # from a JSON string
 reloaded = dp.ProfileReport.from_dict(report.to_dict())  # from a dict
 ```
 
-**Rounding:** All floating-point values in exported data are rounded to match the CLI
-JSON output -- 2 decimal places for percentages and ratios, 4 decimal places for
-statistical metrics. Raw property access on `ColumnProfile` returns unrounded Rust
-values; use the export methods for clean output.
+**Rounding:** All floating-point values in exported data are rounded -- 2 decimal
+places for `0..100` percentages, 4 decimal places for statistical metrics and for
+`0..1` ratios such as `uniqueness_ratio` and `true_ratio`, so that a ratio carries
+the same resolution as the equivalent percentage. Raw property access on
+`ColumnProfile` returns unrounded Rust values; use the export methods for clean
+output.
+
+**Round-trip fidelity:** a report reloaded with `from_dict`, `from_json`, or
+`load` reports the same values as the report it was saved from, at the precision
+above. Reloaded reports are read-only proxies rather than native reports, so the
+two are different objects with the same answers.
 
 ## `ColumnProfile`
 
--- a/python/dataprof/__init__.py
+++ b/python/dataprof/__init__.py
@@ -574,8 +574,9 @@ def column_to_dict(col: ColumnProfile) -> dict[str, Any]:
           "patterns": [{"name": ..., "regex": ..., ...}, ...]
         }
 
-    Floating-point values are rounded to match the Rust serialization
-    (2dp for percentages, 4dp for statistics).
+    Floating-point values are rounded: 2dp for ``0..100`` percentages, 4dp for
+    statistics and for ``0..1`` ratios such as ``uniqueness_ratio``, so that a
+    ratio carries the same resolution as the equivalent percentage.
     """
     col_data: dict[str, Any] = {
         "name": col.name,
@@ -1940,9 +1941,9 @@ class ProfileReport:
     def to_dict(self) -> dict:
         """Convert the report to a nested Python dict.
 
-        All floating-point values are rounded (2dp for percentages, 4dp for
-        statistics) to match the Rust report serialization. The document
-        carries ``schema_version`` (``dataprof.REPORT_SCHEMA_VERSION``) so
+        All floating-point values are rounded: 2dp for ``0..100`` percentages,
+        4dp for statistics and for ``0..1`` ratios such as ``uniqueness_ratio``.
+        The document carries ``schema_version`` (``dataprof.REPORT_SCHEMA_VERSION``) so
         saved reports remain readable across releases; see
         :meth:`from_dict` for the compatibility policy.
         """
--- a/python/dataprof/__init__.py
+++ b/python/dataprof/__init__.py
@@ -232,14 +232,18 @@ def _half_up(v: float, ndigits: int) -> float:
 
 
 def _r2(v: float | None) -> float | None:
-    """Round to 2 decimal places (percentages, ratios). None/NaN → None."""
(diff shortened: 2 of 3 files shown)


AndreaBozzo:
Both Copilot comments were valid and are fixed in 762e669.

**Rounding tie-breaking.** Correct catch, and it was a real defect rather than a wording issue: `_round_candidates` documented its second candidate as reproducing Rust's `(v * 10^n).round() / 10^n` but built it with Python's `round()`, which breaks ties to even where `f64::round()` breaks them away from zero — exactly the tie cases that candidate exists to cover. Verified the disagreement is real before and after:

| value | dp | old `round()` | Rust `f64::round()` |
| --- | --- | --- | --- |
| 0.125 | 2 | 0.12 | **0.13** |
| 0.00125 | 4 | 0.0012 | **0.0013** |

Now quantizes `Decimal(scaled)` — the exact binary value — half-up, which reproduces `f64::round()` faithfully, with an `InvalidOperation` fallback for magnitudes past the decimal context's precision (the same case `_half_up` guards).

**Stale docstring.** Fixed, and the same stale sentence was in `ProfileReport.to_dict()` too, so both are updated. Worth noting the assumption that docstring encoded — "2dp for percentages, 4dp for statistics", with ratios silently filed under percentages — is what produced the `uniqueness_ratio` bug this PR fixes.

Gates re-run: 688 passed, 5 skipped, 1 xfailed; ruff and ty clean.

PR #529, opened 2026-08-03
feat(python): pin the public surface to what __all__ declares

### LINKED ISSUE #514: Pin the Python public surface: __all__ declares 28 names, 40 are reachable
Parent: #511

## Problem

`dataprof.__all__` declares 28 names. 40 are reachable:

```
Any, Callable, Iterator, TYPE_CHECKING, annotations, cast,
column_to_dict, functools, json, math, os, pathlib, warnings
```

Stdlib modules and typing helpers are importable as package attributes
(`dataprof.os`, `dataprof.json`, `dataprof.pathlib`) purely because the whole
package lives in one module namespace. `column_to_dict` is the opposite problem:
the `.pyi` stub documents it, `__all__` omits it, so its status is undecided.

The Rust facade does not have this problem, because `tests/public_api_facade.rs`
pins its surface and `docs/architecture/public-api-inventory.md` records it —
and the inventory currently has **zero drift** across 81 names. Python has no
equivalent.

## Outcome

The Python public surface is what `__all__` says it is, and additions to it are
deliberate.

## Work

- Add a test asserting the non-underscore names reachable on `dataprof` equal
  `__all__`, modelled on `tests/public_api_facade.rs`.
- Decide `column_to_dict`: export it or make it private. The stub and `__all__`
  must agree either way.
- Keep incidental imports out of the package namespace. Once #511's split lands
  this is structural; until then it can be done with explicit `del` or by
  importing under private names.
- Extend the check to `dataprof.agent` and `dataprof.asyncio`, which have the
  same exposure.

## Acceptance criteria

- [ ] Reachable public names equal `__all__` for `dataprof`, `dataprof.agent`,
      and `dataprof.asyncio`.
- [ ] `column_to_dict` has a decided, consistent status in both the stub and
      `__all__`.
- [ ] The test fails when a new incidental import is added.
- [ ] No currently documented public name is removed without a release note.

## Non-goals

- Changing what any public function does.

AndreaBozzo:
Closes #514. Last of the three guards in #511 and a decent showcase of stacked PR new Gh feature.

> **Stacked on #528** — base is `fix/rounding-parity-513`, so the diff here is
> only the surface work. Merge #528 first and this retargets to `master`
> automatically.

## What was leaking

`dataprof.__all__` declared 28 names; 40 were reachable. Because the whole
package lives in one module namespace, every plain `import` landed in it:

```python
>>> import dataprof
>>> dataprof.os, dataprof.json, dataprof.pathlib   # all worked
>>> dataprof.Any, dataprof.cast, dataprof.TYPE_CHECKING
```

It was not just the top-level package:

| module | `__all__` | reachable | leaked |
| --- | --- | --- | --- |
| `dataprof` | 28 | 40 | 13 |
| `dataprof.agent` | 6 | 19 | 13 |
| `dataprof.asyncio` | **none at all** | 12 | 12 |
| `dataprof.interop` | 11 | 13 | 2 |

`dataprof.agent` leaked `ProfileReport`, `StopCondition` and `StructureReport`
on top of the stdlib and typing names — re-exports of public types that were
never meant to be part of *that* module's API. `dataprof.asyncio` had no
`__all__`, so `from dataprof.asyncio import *` pulled in `Path`, `Any` and four
report types alongside the five real functions.

All four now report zero drift.

## How

Everything imported for internal use is bound under a private alias — the
convention the module **already** used for `csv`, `io`, `decimal`, `html`,
`errno` and `importlib`. `os`, `json`, `math`, `functools`, `warnings` and
`pathlib` were simply the ones that got missed.

Renames were done with `tokenize`, not text substitution, so strings and
comments were never touched. That left exactly one thing a tokenizer cannot
see, which the type checker caught:

```python
_cast("str | os.PathLike[str]", roots)   # a type expression inside a string
```

**On the typing names:** I first tried `del Any, Callable, Iterator` at the end
of the module — the approach #514 suggests. It does not work: ruff and ty read
a module-scope `del` as an unconditional unbind and flagged every annotation in
the file as `Undefined name` (41 errors). Private aliases it is.

## `column_to_dict` is public

It was in the `.pyi` stub, announced in the changelog, and used in
`docs/guides/examples.md:414` — public in every way except `__all__`. Making it
private would have removed a documented name, so it is exported. Two existing
tests encoded the old undecided state, including one literally named
`test_column_to_dict_discoverable_from_interop_not_all`; both are updated.

## The guard

Two overlapping checks, modelled on `tests/public_api_facade.rs`:

- **structural** — reachable names equal `__all__`. Catches a new incidental
  import with nobody editing a list.
- **inventory** — the literal expected names. The structural check would happily
  accept a *new public name*; this one makes growing the API a deliberate edit
  that shows up in review.

Package submodules are ignored: importing `dataprof.interop` anywhere binds
`interop` on the parent by design, and counting it made the check depend on test
ordering — which is how it passed alone and failed in the full suite.

`docs/architecture/public-api-inventory.md` gains a Python section, so the two
languages are recorded the same way.

## Verification

Both halves sabotaged, with targeted reverts:

| Sabotage | Caught by |
| --- | --- |
| `import textwrap` added to `__init__.py` | structural: `reachable but not declared: ['textwrap']` |
| new export `profile_everything` | inventory + star-import: `added: ['profile_everything']` |

```
uv run pytest python/tests/ -q     # 711 passed, 5 skipped, 1 xfailed
uv run ruff format/check, uv run ty check python/
cargo test --test public_api_facade # 15 passed (Rust facade unaffected)
```

No public name is removed, so no release note is needed on that count;
`column_to_dict` is *added* to the declared surface.

### DIFF
--- a/docs/architecture/public-api-inventory.md
+++ b/docs/architecture/public-api-inventory.md
@@ -68,3 +68,30 @@ CI now enforces the lean facade combinations with `cargo check`,
 `public_api_facade` integration-test runs, and owning-crate unit tests for
 `dataprof-engines`, `dataprof-parquet`, and `dataprof-partial` so the facade
 cannot accidentally grow implementation ownership again.
+
+## Python Package Surface
+
+The Python package is held to the same rule as the Rust facade: what
+`__all__` declares is what is reachable. `python/tests/test_public_surface.py`
+enforces it for every module below, and carries the same names as
+`EXPECTED_SURFACE`, so growing the API is a deliberate edit rather than a side
+effect.
+
+| Module | Exports |
+| --- | --- |
+| `dataprof` | `Capabilities`, `capabilities`, `REPORT_SCHEMA_VERSION`, `profile`, `profile_file`, `Profiler`, `ProfileReport`, `ProfilerConfig`, `ColumnProfile`, `DataQualityMetrics`, `SamplingStrategy`, `StopCondition`, `ProgressEvent`, `list_patterns`, `infer_schema`, `quick_row_count`, `analyze_structure`, `SchemaResult`, `RowCountEstimate`, `StructureColumnSummary`, `StructureReport`, `RecordBatch`, `column_to_dict`, `asyncio`, `__version__`, `analyze_database_async`, `count_table_rows_async`, `get_table_schema_async`, `test_connection_async` |
+| `dataprof.agent` | `AgentGuard`, `AgentSecurityError`, `AgentTimeoutError`, `PathNotAllowedError`, `ResourceLimitExceededError`, `SandboxPolicy` |
+| `dataprof.asyncio` | `profile_bytes`, `profile_file`, `profile_url`, `infer_schema_stream`, `quick_row_count_stream` |
+| `dataprof.interop` | `analyze_file`, `profile_dataframe`, `profile_arrow`, `analyze_csv_to_arrow`, `analyze_parquet_to_arrow`, `column_to_dict`, `ProfilerConfig`, `ProfileReport`, `ColumnProfile`, `DataQualityMetrics`, `RecordBatch` |
+
+### Not Public
+
+Everything the implementation imports for its own use is bound under a private
+alias — `import os as _os`, `from typing import Any as _Any` — so it does not
+land in the package namespace. Before this was enforced, `dataprof.os`,
+`dataprof.json` and `dataprof.pathlib` were all importable, and
+`dataprof.asyncio` declared no `__all__` at all.
+
+The package's own submodules (`dataprof.agent`, `dataprof.interop`) become
+attributes of `dataprof` once imported anywhere in the process. That is normal
+package behaviour rather than a leak, and the guard ignores it.
--- a/python/dataprof/__init__.py
+++ b/python/dataprof/__init__.py
@@ -1,30 +1,30 @@
 """dataprof - High-performance data profiling library."""
 
-from __future__ import annotations
+from __future__ import annotations as _annotations
 
 import csv as _csv
 import decimal as _decimal
 import errno as _errno
-import functools
+import functools as _functools
 import html as _html
 import importlib as _importlib
 import io as _io
-import json
-import math
-import os
-import pathlib
-import warnings
-from collections.abc import Callable, Iterator
+import json as _json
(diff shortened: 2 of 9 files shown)


AndreaBozzo:
Checked this before writing it, and re-verified now: **aliasing a future import is not a SyntaxError**, and the plain form is what causes the leak this PR removes.

```
1. Is aliasing a future import a SyntaxError?
   NO - compiles fine

2. Does the plain form bind a public `annotations` name?
   'annotations' in namespace: True  -> _Feature((3, 7, 0, 'beta', 1), None, 16777216)

3. Does the ALIASED form still enable PEP 563?
   plain  : __annotations__ = {'x': 'Undefined', 'return': 'Nope'}
   aliased: __annotations__ = {'x': 'Undefined', 'return': 'Nope'}
   neither: NameError -> PEP 563 NOT active     <- control, proves the check discriminates

4. On the shipped package:
   hasattr(dataprof, 'annotations'): False
   profile_file annotations are strings: True
   dataprof / .agent / .asyncio / .interop -> PEP 563 flag set in compiled module: True
```

The grammar for a future statement is `feature ["as" identifier]` — the alias is permitted, and the compiler enables the feature from the statement's *presence*, not from the name it binds. All 715 tests pass, which they could not if any of these four modules raised `SyntaxError` on import.

The second claim — *"it does not bind a public `annotations` name anyway"* — is the inverse of the truth, and is why the alias is there: the unaliased form binds a `__future__._Feature` object as `dataprof.annotations`, one of the 13 leaked names this PR removes.

Since the objection is plausible enough to come up again, I've pinned the invariant instead of just arguing it (78ef880): a test asserting PEP 563 is active in all four modules **and** that `annotations` is unreachable. Reverting one module to the unaliased form fails both it and the structural check:

```
AssertionError: dataprof.interop: reachable but not declared in __all__: ['annotations']
AssertionError: dataprof.interop: `annotations` is reachable, so the future import is unaliased again
```

No change to the future imports themselves.

AndreaBozzo:
Temporarily closing to dissolve the GitHub stack: stacked PRs can only be merged via the async endpoint, which has no admin bypass and so cannot satisfy the required-review rule. Reopening against master right after #528 lands.

PR #535, opened 2026-08-06
docs: add recipes for validating saved ProfileReport JSON

### LINKED ISSUE #500: Docs: add a CI recipe for validating saved ProfileReport JSON
## User problem

#413 published a versioned JSON Schema for serialized `ProfileReport` values, but the user documentation only links to the artifact. A team that wants to reject malformed or incompatible saved reports still has to invent the validation command and CI wiring.

Add a copy-pasteable recipe that turns the published schema into a practical consumer-side check.

## Scope

Document two small validation paths:

1. a local Python example using the `jsonschema` package
2. a CI-friendly command or GitHub Actions snippet that validates one or more saved `report.json` files against `profile-report.v1.schema.json`

Use the stable published schema URL for downstream consumers and show the repository-local schema path for offline/project CI. Include one deliberately invalid field example so users can see the failure is meaningful.

Likely documentation:

- `docs/guides/examples.md`
- `docs/schema/README.md`
- optionally a small focused example under an examples/docs-support directory if that keeps the recipe executable

## Acceptance criteria

- [ ] A fresh `dp.profile(...).save("report.json")` document passes the recipe.
- [ ] A report with an invalid `schema_version`, enum, or primitive type fails.
- [ ] The example uses JSON Schema Draft 2020-12.
- [ ] The offline variant performs no network fetches.
- [ ] The recipe explains that schema versions are independent of package versions.
- [ ] No new runtime dependency is added to the base `dataprof` wheel.
- [ ] Commands are exercised locally or in a focused documentation test.

## Non-goals

- adding a new validation method to `ProfileReport`
- changing the generated schema or compatibility policy
- validating arbitrary user datasets rather than saved dataprof reports

aryansk:
Summary

Adds the CI and local validation recipes requested by #500 for saved ProfileReport JSON.

Changes

- Document the stable ProfileReport v1 JSON Schema URL.
- Add an online validation recipe using Draft 2020-12 and an offline recipe for projects that vendor the schema.
- Include a deliberately invalid example and a GitHub Actions recipe that validates report artifacts.
- Keep jsonschema as an optional validation-tool dependency rather than adding it to the dataprof runtime requirements.

Validation

- Valid and invalid examples were exercised with the repository development environment.
- Embedded Python recipes parse successfully with the Python AST parser.
- git diff --check passed.

Closes #500

### DIFF
--- a/docs/schema/README.md
+++ b/docs/schema/README.md
@@ -42,3 +42,120 @@ When changing a serialized report field, review compatibility first. Additive
 fields with reader defaults stay in the current schema version; an incompatible
 change requires incrementing `REPORT_SCHEMA_VERSION` and committing a new
 versioned schema without deleting schemas that supported releases still read.
+
+## Validate a saved report
+
+The schema is also useful to consumers that store reports outside dataprof. The
+published URL is stable across package releases:
+
+```text
+https://andreabozzo.github.io/dataprof/schema/profile-report.v1.schema.json
+```
+
+For example, create a full JSON report and validate it with the optional
+`jsonschema` package. The package is a validation-tool dependency only; it is
+not required by the base `dataprof` wheel.
+
+```bash
+python -m pip install jsonschema
+python - <<'PY'
+import copy
+import json
+import os
+from pathlib import Path
+from urllib.request import urlopen
+
+from jsonschema import Draft202012Validator
+
+SCHEMA_URL = (
+    "https://andreabozzo.github.io/dataprof/schema/"
+    "profile-report.v1.schema.json"
+)
+REPORT_PATH = Path("report.json")
+schema_source = os.environ.get("DATAPROF_SCHEMA", SCHEMA_URL)
+
+if schema_source.startswith(("http://", "https://")):
+    with urlopen(schema_source) as response:  # noqa: S310 - URL is explicit above
+        schema = json.load(response)
+else:
+    schema = json.loads(Path(schema_source).read_text(encoding="utf-8"))
+
+Draft202012Validator.check_schema(schema)
+report = json.loads(REPORT_PATH.read_text(encoding="utf-8"))
+validator = Draft202012Validator(schema)
+errors = sorted(validator.iter_errors(report), key=lambda error: list(error.path))
+if errors:
+    for error in errors:
+        print(f"{REPORT_PATH}: {error.message}")
+    raise SystemExit(1)
+print(f"{REPORT_PATH}: valid")
+
+# A deliberately invalid document fails for a useful reason, rather than
+# merely failing because the file could not be read.
+invalid_report = copy.deepcopy(report)
+invalid_report["schema_version"] = "not-an-integer"
+invalid_errors = list(validator.iter_errors(invalid_report))
+assert invalid_errors, "the invalid schema_version should be rejected"
+print(f"invalid example: rejected ({invalid_errors[0].message})")
+PY
+```
+
+The `report.json` input above can be produced by the normal Python API:
+
+```python
+import dataprof as dp
+
+dp.profile("data.csv").save("report.json")
+```
+
+For offline or project CI validation, save the Python block above as
+`validate_report.py` and use the checked-in artifact instead of the network URL:
+
+```bash
+DATAPROF_SCHEMA=docs/schema/profile-report.v1.schema.json python validate_report.py
+```
+
+A minimal GitHub Actions job can validate every report committed under
+`reports/` without adding anything to the runtime package:
+
+```yaml
+name: Validate dataprof reports
+
+on: [push, pull_request]
+
+jobs:
+  reports:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-python@v5
+        with:
+          python-version: "3.x"
+      - run: python -m pip install jsonschema
+      - name: Validate saved reports
+        run: |
+          python - <<'PY'
+          import json
+          from pathlib import Path
+
+          from jsonschema import Draft202012Validator
+
+          schema_path = Path("docs/schema/profile-report.v1.schema.json")
+          schema = json.loads(schema_path.read_text(encoding="utf-8"))
+          validator = Draft202012Validator(schema)
+          report_paths = sorted(Path("reports").rglob("*.json"))
+          if not report_paths:
+              raise SystemExit("no JSON reports found under reports/")
+
+          failures = []
+          for report_path in report_paths:
+              report = json.loads(report_path.read_text(encoding="utf-8"))
+              failures.extend(
+                  f"{report_path}: {error.message}"
+                  for error in validator.iter_errors(report)
+              )
+          if failures:
+              raise SystemExit("\n".join(failures))
+          print(f"validated {len(report_paths)} report(s)")
+          PY
+```
--- a/docs/schema/README.md
+++ b/docs/schema/README.md
@@ -56,6 +56,13 @@ For example, create a full JSON report and validate it with the optional
 `jsonschema` package. The package is a validation-tool dependency only; it is
 not required by the base `dataprof` wheel.
 
+Because the schema allows unknown additive properties, a successful validation
+confirms required fields and primitive types but does not reject extra keys.
+The Python API writes the `PythonProfileReportDocument` shape, so the examples
+below pin validation to that branch of the versioned schema. This keeps errors
+focused on the fields that matter to Python consumers instead of reporting the
+whole document as failing a top-level `anyOf`.
+
 ```bash
 python -m pip install jsonschema
 python - <<'PY'
@@ -75,18 +82,23 @@ REPORT_PATH = Path("report.json")
 schema_source = os.environ.get("DATAPROF_SCHEMA", SCHEMA_URL)
 
 if schema_source.startswith(("http://", "https://")):
-    with urlopen(schema_source) as response:  # noqa: S310 - URL is explicit above
+    with urlopen(schema_source, timeout=30) as response:  # noqa: S310 - project default or caller-supplied URL
         schema = json.load(response)
 else:
     schema = json.loads(Path(schema_source).read_text(encoding="utf-8"))
 
-Draft202012Validator.check_schema(schema)
+document_schema = {key: value for key, value in schema.items() if key != "anyOf"}
+document_schema["$ref"] = "#/$defs/PythonProfileReportDocument"
+Draft202012Validator.check_schema(document_schema)
 report = json.loads(REPORT_PATH.read_text(encoding="utf-8"))
-validator = Draft202012Validator(schema)
-errors = sorted(validator.iter_errors(report), key=lambda error: list(error.path))
+validator = Draft202012Validator(document_schema)
+errors = sorted(
+    validator.iter_errors(report),
+    key=lambda error: [str(path_part) for path_part in error.path],
+)
 if errors:
     for error in errors:
-        print(f"{REPORT_PATH}: {error.message}")
+        print(f"{REPORT_PATH}: {error.json_path}: {error.message}")
     raise SystemExit(1)
 print(f"{REPORT_PATH}: valid")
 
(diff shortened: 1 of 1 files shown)


AndreaBozzo:
Hi @aryansk and thanks for your contribution on dataprof, please set the PR as ready for review when done so i'll take a look at it

aryansk:
Thanks for the guidance. The Apply-path and Semgrep checks are passing, so I’ve marked the PR ready for review.

docs/schema/README.md
```diff
@@ -42,3 +42,120 @@ When changing a serialized report field, review compatibility first. Additive
...
+
+if schema_source.startswith(("http://", "https://")):
+    with urlopen(schema_source) as response:  # noqa: S310 - URL is explicit above
+        schema = json.load(response)
+else:
+    schema = json.loads(Path(schema_source).read_text(encoding="utf-8"))
+
+Draft202012Validator.check_schema(schema)
+report = json.loads(REPORT_PATH.read_text(encoding="utf-8"))
+validator = Draft202012Validator(schema)
```
AndreaBozzo: This is the main thing worth changing.

The root of the schema is an `anyOf` over two branches, `ProfileReport` (the Rust shape) and `PythonProfileReportDocument` (what `dp.profile(...).save()` writes). Validating against the root means every failure is reported as "not valid under any of the given schemas", and jsonschema puts the **entire instance** in `error.message`.

Measured on a 3.5 KB toy report:

| mutation | `error.message` length |
| --- | --- |
| `schema_version` to a string | 2700 chars |
| drop `quality` | 1469 chars |
| `execution.rows_processed` to a string | 2687 chars |

All three are the whole report dumped back at you, with no indication of which field was wrong. On a real report with a few hundred columns this is enormous, and the CI snippet prints one per error per file.

Pinning the root to the branch the document actually is fixes it. A report saved by the Python API is a `PythonProfileReportDocument`:

```python
doc_schema = {k: v for k, v in schema.items() if k != "anyOf"}
doc_schema["$ref"] = "#/$defs/PythonProfileReportDocument"
validator = Draft202012Validator(doc_schema)
```

Same three mutations, verified:

```
$.schema_version: 'not-an-integer' is not of type 'integer'
$: 'quality' is a required property
$.execution.rows_processed: 'x' is not of type 'integer'
```

That is the meaningful failure #500 asked for. It is worth a sentence of prose next to it explaining that the schema covers two document shapes and which one the Python API produces, since that is not obvious and it is what makes the `$ref` line make sense.

docs/schema/README.md
```diff
@@ -42,3 +42,120 @@ When changing a serialized report field, review compatibility first. Additive
...
+    raise SystemExit(1)
+print(f"{REPORT_PATH}: valid")
+
+# A deliberately invalid document fails for a useful reason, rather than
+# merely failing because the file could not be read.
+invalid_report = copy.deepcopy(report)
+invalid_report["schema_version"] = "not-an-integer"
+invalid_errors = list(validator.iter_errors(invalid_report))
+assert invalid_errors, "the invalid schema_version should be rejected"
+print(f"invalid example: rejected ({invalid_errors[0].message})")
```
AndreaBozzo: This line is where the message problem is most visible, because it is the one the reader is meant to look at. As written it prints:

```
invalid example: rejected ({'schema_version': 'not-an-integer', 'source': 'C:\Users\...', ... 2700 more chars})
```

The comment above it says the document "fails for a useful reason", but the output does not show a reason at all. With the `$ref` change suggested on the validator line it becomes:

```
invalid example: rejected ($.schema_version: 'not-an-integer' is not of type 'integer')
```

Printing `f"{error.json_path}: {error.message}"` rather than `error.message` alone is worth doing in the two loops above as well, since the path is what makes a failure actionable when a report has many columns.

docs/schema/README.md
```diff
@@ -42,3 +42,120 @@ When changing a serialized report field, review compatibility first. Additive
...
+if schema_source.startswith(("http://", "https://")):
+    with urlopen(schema_source) as response:  # noqa: S310 - URL is explicit above
+        schema = json.load(response)
+else:
+    schema = json.loads(Path(schema_source).read_text(encoding="utf-8"))
+
+Draft202012Validator.check_schema(schema)
+report = json.loads(REPORT_PATH.read_text(encoding="utf-8"))
+validator = Draft202012Validator(schema)
+errors = sorted(validator.iter_errors(report), key=lambda error: list(error.path))
```
AndreaBozzo: On the existing Copilot comment about this sort key raising `TypeError`: applying it is harmless, but the concern does not hold for this schema, so please do not treat it as a bug you have to fix.

`error.path` is an instance path, and any single parent in the document is either an array or an object, never both. Two errors therefore cannot differ by `int` versus `str` at the same position. On top of that, with the root `anyOf` in place every top-level error has an empty path, so the key is comparing `[]` against `[]`.

If you take the `$ref` change suggested above, paths become non-empty and meaningful, and `key=lambda error: [str(p) for p in error.path]` is then a reasonable defensive choice. Just be aware it orders index 10 before index 9.

docs/schema/README.md
```diff
@@ -42,3 +42,120 @@ When changing a serialized report field, review compatibility first. Additive
...
+        run: |
+          python - <<'PY'
+          import json
+          from pathlib import Path
+
+          from jsonschema import Draft202012Validator
+
+          schema_path = Path("docs/schema/profile-report.v1.schema.json")
+          schema = json.loads(schema_path.read_text(encoding="utf-8"))
+          validator = Draft202012Validator(schema)
```
AndreaBozzo: Same `anyOf` issue as the first recipe, and it matters more here. `failures` collects one message per error per file, and each message is the full JSON document, so a CI job validating a directory of reports fails with megabytes of dumped JSON and no indication of which field is wrong in which file.

The same two-line `$ref` fix applies. Including `error.json_path` in the failure string is also worth it here:

```python
f"{report_path}: {error.json_path}: {error.message}"
```

The Copilot suggestion to call `Draft202012Validator.check_schema(schema)` here too is worth taking. It is one line, it matches what the first recipe already does, and it gives a clear failure if the vendored copy has been edited.

docs/schema/README.md
```diff
@@ -42,3 +42,120 @@ When changing a serialized report field, review compatibility first. Additive
...
+
+SCHEMA_URL = (
+    "https://andreabozzo.github.io/dataprof/schema/"
+    "profile-report.v1.schema.json"
+)
+REPORT_PATH = Path("report.json")
+schema_source = os.environ.get("DATAPROF_SCHEMA", SCHEMA_URL)
+
+if schema_source.startswith(("http://", "https://")):
+    with urlopen(schema_source) as response:  # noqa: S310 - URL is explicit above
```
AndreaBozzo: Endorsing the existing Copilot comment on this line: the timeout is worth adding. `urlopen` with no timeout can hang a CI job indefinitely, and this snippet is explicitly offered as a CI recipe.

```python
with urlopen(schema_source, timeout=30) as response:  # noqa: S310
```

The `noqa` justification is also slightly off now that `DATAPROF_SCHEMA` can point anywhere, so trimming the trailing explanation or rewording it to "URL is the project default or explicitly supplied by the caller" would be accurate.

docs/schema/README.md
```diff
@@ -42,3 +42,120 @@ When changing a serialized report field, review compatibility first. Additive
...
+
+Draft202012Validator.check_schema(schema)
+report = json.loads(REPORT_PATH.read_text(encoding="utf-8"))
+validator = Draft202012Validator(schema)
+errors = sorted(validator.iter_errors(report), key=lambda error: list(error.path))
+if errors:
+    for error in errors:
+        print(f"{REPORT_PATH}: {error.message}")
+    raise SystemExit(1)
+print(f"{REPORT_PATH}: valid")
```
AndreaBozzo, line 103: Minor, and purely a docs sentence rather than a code change.

`additionalProperties` is not set to `false` anywhere in the schema, so an unknown or misspelled top-level key passes validation. I checked: adding `{"totally_bogus": {...}}` to a valid report is accepted.

That is normal JSON Schema behaviour and not something to fix in this PR, but a reader who has just been shown a validation recipe will reasonably assume a passing check means the document is exactly right. One sentence saying the check catches missing and wrong-typed fields, not unexpected extra ones, would set the expectation correctly.

AndreaBozzo (commented):
Thanks for this, and welcome. The structure is right, the acceptance criteria in #500 are addressed one by one, and keeping `jsonschema` out of the runtime dependencies is the correct call.

I ran the recipe rather than only reading it, against a report saved from current `master` (`f543224`), which includes this week's parser changes. What holds up:

* A fresh `dp.profile(...).save("report.json")` validates clean.
* The published URL returns HTTP 200 and is byte-for-byte the same schema as the checked-in copy once line endings are normalised, so the online and offline recipes genuinely agree. Worth knowing, because that is the kind of thing that silently drifts.
* `schema_version = "not-an-integer"` really is rejected: the schema pins it to `const: 1`, `type: integer`.
* Corruptions of real fields are caught. I tried dropping `quality`, `columns` as a string, and `execution.rows_processed` as a string; all three fail as they should.
* The heredoc in the Actions snippet is correct. YAML block-scalar de-indentation puts the `PY` terminator at column 0, which is the thing that usually breaks in these.

One substantive problem, which I do not think shows up from reading the diff. The error messages the recipe prints are unusable, so the part of #500 asking that "users can see the failure is meaningful" is not met yet. Details inline on the two places it bites, with a fix I verified.

Beyond that, two small documentation gaps and a note on the existing Copilot comments, also inline. Nothing here needs a redesign; the recipe is close.

aryansk:
Thanks for the detailed review. I addressed the actionable comments in commit 7da869b:

- Both examples now pin validation to the PythonProfileReportDocument branch instead of validating the top-level anyOf as a whole.
- Errors are sorted by path and include json_path plus the message, so each failure points to the field that needs attention.
- Both examples run Draft202012Validator.check_schema before validating reports.
- The URL example uses a 30 second timeout, and the S310 note now covers the project default or a caller-supplied URL.
- The README explains that additive properties are allowed while required fields and primitive types are still checked.

Validation: cargo test --test profile_report_schema passed 7/7; the Python validation example passed with a focused $.schema_version error; git diff --check passed. The PR remains ready for review.

AndreaBozzo (commented):
All six points addressed. I re-ran both recipes verbatim from this commit rather than reading them, extracting the Python straight out of the markdown so what I executed is exactly what a reader would copy.

Local recipe, against a report saved from current `master`:

```
report.json: valid
invalid example: rejected ($.schema_version: 1 was expected)
```

CI recipe, against a directory holding one good report and one with `execution.rows_processed` corrupted to a string:

```
validated 1 report(s)
reports/bad.json: $.execution.rows_processed: 'x' is not of type 'integer'
```

That is the behaviour #500 asked for. The failure now names the file, the field and the problem, where before it printed 2700 characters of dumped JSON.

The prose you added is the right call too. Saying up front that the Python API writes the `PythonProfileReportDocument` shape, and that a pass confirms required fields and types without rejecting extra keys, sets the expectation correctly before anyone copies the block.

One optional nit, not worth another round on its own. `invalid_errors[0]` happens to surface the `const` failure ("1 was expected") rather than the type failure, because both fire for that mutation. It is still actionable, so only worth changing if you touch the file again.

Good to merge from my side.

Separately, and not something for you to fix here: this PR runs only two checks, because `ci.yml` and `security.yml` both carry `paths-ignore: docs/**`. That is deliberate and correct for documentation, but it does mean nothing in CI exercises these recipes, which is why I ran them by hand twice. I will open a follow-up about that.

AndreaBozzo (approved):

aryansk:
Thanks for rerunning both recipes from the actual markdown and for confirming all six review points. I really appreciate the careful validation and the thoughtful review. Glad to see this land in master.

AndreaBozzo:
Thanks to you and your support, leave a ⭐ to dataprof if you can

Take care,
Andrea

aryansk:
Thank you, Andrea. I am glad I could help. Take care!

PR #556, opened 2026-08-09
fix(python): make llm context null flags round-trip stable

### LINKED ISSUE #526: to_llm_context flags threshold on unrounded values, so a round-trip changes the flags

aryansk:
## Summary

- Compare null-heavy thresholds against the same two-decimal percentage used in serialized reports, so `to_llm_context()` stays stable across round-trips.
- Use exact null and total counts to distinguish genuinely all-null columns from values whose percentage rounds to `100.0%`.
- Replace the strict round-trip xfail with a passing parity regression and add an exact-count boundary test.

Closes #526

## Validation

- `uv run pytest python/tests/test_report_roundtrip_parity.py -q` — 20 passed
- `uv run pytest python/tests/ -q` — 825 passed, 107 skipped
- `uv run ruff format --check python/dataprof/__init__.py python/tests/test_report_roundtrip_parity.py`
- `uv run ruff check python/dataprof/__init__.py python/tests/test_report_roundtrip_parity.py`
- `uv run ty check python/`

### DIFF
--- a/python/dataprof/__init__.py
+++ b/python/dataprof/__init__.py
@@ -939,21 +939,27 @@ def _column_flags(col: ColumnProfile) -> list[tuple[float, str]]:
     emitted -- a flag that fires on almost every column is noise, not signal.
     """
     flags: list[tuple[float, str]] = []
-    null_pct = col.null_percentage or 0.0
+    rounded_null_pct = _r2(col.null_percentage)
+    null_pct_for_flags = rounded_null_pct if rounded_null_pct is not None else 0.0
+    total = col.total_count or 0
     name = _one_line(col.name)
 
-    if null_pct >= 100.0:
+    # Do not infer "all-null" from the rounded percentage: a non-null value
+    # can legitimately round 99.996% up to 100.0%. Counts preserve the exact
+    # structural distinction, while the threshold and rendered percentage use
+    # the same serialized precision so native and round-tripped reports agree.
+    is_all_null = total > 0 and col.null_count is not None and col.null_count == total
+    if is_all_null:
         flags.append((100.0, f"{name}: all-null"))
-    elif null_pct >= _NULL_HEAVY_PCT:
-        flags.append((null_pct, f"{name}: null-heavy ({_r2(null_pct):.1f}% null)"))
+    elif rounded_null_pct is not None and rounded_null_pct >= _NULL_HEAVY_PCT:
+        flags.append((rounded_null_pct, f"{name}: null-heavy ({rounded_null_pct:.1f}% null)"))
 
     # A single distinct value carries no information. Suppress when the column
     # is already reported as null-heavy, where it is a restatement, not a flag.
-    if col.unique_count == 1 and null_pct < _NULL_HEAVY_PCT and (col.total_count or 0) > 1:
+    if col.unique_count == 1 and null_pct_for_flags < _NULL_HEAVY_PCT and total > 1:
         flags.append((60.0, f"{name}: constant (1 distinct value)"))
 
     outliers = col.outlier_count or 0
-    total = col.total_count or 0
     if outliers > 0 and total > 0:
         pct = 100.0 * outliers / total
         flags.append((min(50.0, pct), f"{name}: {outliers} outliers ({pct:.1f}%)"))
--- a/python/tests/test_report_roundtrip_parity.py
+++ b/python/tests/test_report_roundtrip_parity.py
@@ -27,10 +27,12 @@ import json
 import math
 import warnings
 from decimal import ROUND_HALF_UP, Decimal, InvalidOperation
+from types import SimpleNamespace
 from typing import Any
 
 import dataprof
 import pytest
+from dataprof import _column_flags  # ty: ignore[unresolved-import]
 
 # Members that cannot participate in a value-parity sweep, each for a reason
 # that is not "it disagrees":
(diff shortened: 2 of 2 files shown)


AndreaBozzo:
Thank you @aryansk , i'll review this once is set to ready

aryansk:
Thanks, Andrea — I’ve marked the PR ready for review now that the hosted checks are green. The focused and full test suites, Ruff, ty, and the hosted checks are passing. I’ll keep an eye out for your review.

AndreaBozzo (approved):
Approving. I verified the fix beyond the green checks by reverting each half separately and confirming the tests fail:

- Reverting the whole fix: 4 failures (the three `test_report_members_survive_roundtrip` parametrizations plus `test_llm_context_flags_survive_roundtrip`).
- Reverting only the rounded-threshold half: the same 4 failures.
- Reverting only the counts-based `is_all_null` half: `test_all_null_flag_requires_exact_counts` fails.

So both halves are independently guarded, which is what I wanted to see. Worth stating precisely what that last test does, because it is easy to misread as covering a bug on `master`: it does not. On `master` the check compares the raw percentage, so 99.996% never claimed "all-null". The counts check is what keeps the *other* half of this fix from introducing that defect once the threshold moves to the rounded value. That is the right way round, and the test earns its place.

Removal of the `strict=True` xfail and `_KNOWN_DIVERGENT_REPORT_MEMBERS` is complete — definition and both use sites.

Three non-blocking notes, none of which need to hold up the merge:

1. The docstring on `test_llm_context_flags_survive_roundtrip` still describes the bug in the present tense — "native renders it as '20.0% null' without the null-heavy flag". After this change native *does* flag it. It reads as the old xfail reason with only the conclusion swapped, and will mislead the next reader.

2. This changes agent-facing output, and the PR body does not say which direction was chosen. A column at 19.998% null now gets a null-heavy flag it never got before. Reconciling toward "round up into the flag" is the right call — the rendered number and the threshold finally come from the same precision — but that is the user-visible consequence and #526 flagged it as the reason the fix was deferred, so it belongs in the description.

3. `null_count is None` now degrades all-null to null-heavy. `_column_to_dict` always writes `null_count`, so real round-trips are unaffected and only a hand-built `from_dict` payload reaches it. That degradation is consistent with the `None` = "not analyzed" convention, but it is neither tested nor mentioned.

Minor: `rounded_null_pct` and `null_pct_for_flags` are redundant. With `_NULL_HEAVY_PCT = 20.0`, `null_pct_for_flags >= _NULL_HEAVY_PCT` is exactly equivalent to the `is not None and >=` guard on the `elif`, so one variable covers both uses.

Thanks for the careful work on this one — the separated-halves test structure is what made it quick to verify.

PR #557, opened 2026-08-09
fix(partial): preserve header-only CSV schema

### LINKED ISSUE #552: infer_schema reports 0 columns for a header-only CSV; profile and analyze_structure report the header
## Summary

For a CSV with a header and no data rows, `infer_schema` reports zero columns and
an empty name list. `profile` and `analyze_structure` both report the two columns
the header declares. Three public entry points, one file, two different answers.

The empty answer is also reported as `schema_stable=True`, so the API expresses
confidence in it.

## Reproduction

```python
import dataprof as dp
from pathlib import Path

p = Path("hdr_only.csv")
p.write_bytes(b"a,b\n")

pr = dp.profile(p)
sc = dp.infer_schema(p)
st = dp.analyze_structure(p)

print("profile          :", pr.rows, pr.columns, [c.name for c in pr.profiles])
print("infer_schema     :", sc.num_columns, list(sc.column_names), sc.schema_stable)
print("analyze_structure:", [c.name for c in st.columns], st.row_count.count)
```

```
profile          : 0 2 ['a', 'b']
infer_schema     : 0 [] True
analyze_structure: ['a', 'b'] 0
```

## Why it matters

The header is present and unambiguous, so the schema of this file is two columns
named `a` and `b`. Zero columns is not a cheaper approximation of that answer, it
is a different and wrong one.

This also collides with the repo's stated `None`-versus-empty convention. An empty
`column_names` means "analyzed, found nothing", which asserts the file declares no
columns. The file declares two.

The likely path to hitting this is a schema check in a pipeline: an upstream job
writes a header and no rows on a day with no data, and a downstream guard using
`infer_schema` concludes the schema changed or vanished, when nothing about the
schema changed. `profile` on the same file would have told it the truth.

Column count is the one thing `infer_schema` exists to answer cheaply, and a
header-only file is the cheapest case there is.

## Suggested resolution

Derive the columns from the header for delimited formats, independent of how many
data rows follow, which is what `analyze_structure` already does. Types for a
column with no observed values should follow the existing convention for "no
evidence" rather than defaulting to a concrete type.

The genuinely empty cases are already consistent, so this is the only divergence:

```
empty_array.json   profile=(0 rows, 0 cols)  schema=(0, [])  structure=[]
empty.jsonl        profile=(0 rows, 0 cols)  schema=(0, [])  structure=[]
empty.csv          profile=(0 rows, 0 cols)  schema=(0, [])  structure=[]
```

All three agree on zero columns there, which is right: those files declare no
columns. Header-only CSV is the one case where a declaration exists and
`infer_schema` alone ignores it.

- `infer_schema` on a header-only CSV reports the header's columns and names.
- `profile`, `infer_schema`, and `analyze_structure` agree on column names and
  count for the same file, asserted by a test across all supported formats.
- The zero-row case is covered for JSON and JSONL with whatever answer is decided,
  stated explicitly rather than falling out of the implementation.

Found during 0.11 dogfooding of the partial-analysis surface.

## Related

- The companion `schema_stable` issue: the same file reports `schema_stable=True`
  for this empty result.

floze-the-genius:
# Pull Request

## 📝 Summary

Preserve declared CSV columns in partial schema inference when a file or stream contains headers but no data rows. This brings `infer_schema` into line with `profile` and `analyze_structure`: columns without observations use the existing no-evidence `String` type, while empty JSON and JSONL inputs continue to produce zero columns.

## 🔧 Type of Change

- [x] 🐛 Bug fix (non-breaking change that fixes an issue)
- [ ] ✨ New feature (non-breaking change that adds functionality)
- [ ] 🚨 Breaking change (would cause existing functionality to not work as expected)
- [ ] 📚 Documentation update
- [ ] ⚡ Performance improvement
- [ ] 🎨 Code refactoring (no functional changes)

## 📋 Changes Made

- map every declared header into `SchemaResult` instead of dropping columns without streaming statistics
- fall back to the existing `DataType::String` no-evidence convention
- cover sync and async header-only CSV paths plus exact name/type parity across Python `profile`, `infer_schema`, and `analyze_structure`
- preserve zero-column behavior for empty JSON arrays and blank JSONL

**Related Issue:** Closes #552

### How did you test this?

- [x] Tested locally with `cargo test`
- [x] Added new tests for this feature
- [x] All existing tests pass in the affected package and focused Python class
- [x] Tested with sample in-memory CSV/JSON/JSONL data
- [x] Tested empty/header-only edge cases

### Test Commands Run

```bash
cargo test -p dataprof-partial --no-default-features
cargo test -p dataprof-partial --all-features
uv run maturin develop
uv run pytest python/tests/test_python_api.py::TestPartialAnalysis -q
cargo fmt --all -- --check
uv run ruff format --check python/ .github/scripts/ .claude/skills/dataprof/scripts/
uv run ty check python/
git diff --check
```

Results: 31 unit + 2 doc tests pass without default features; 44 unit + 4 doc tests pass with all features; the Python `TestPartialAnalysis` class passes 11/11.

## 📖 Documentation

- [ ] Updated README.md if user-facing changes
- [ ] Added/updated inline code comments
- [ ] Updated docs/ folder if architectural changes
- [x] No documentation changes needed

## ⚡ Performance Impact

- [x] No performance impact
- [ ] Performance improved
- [ ] Performance may be affected (explain below)

## 🚨 Breaking Changes

- [x] No breaking changes

## 📌 Additional Notes

The change is confined to partial schema construction. It does not initialize global parser statistics or alter profile/quality calculations. The direct `analyze_csv_file` parser path is tracked separately in #558.

## ✅ Checklist

- [x] Code builds without warnings
- [x] No linting errors (`cargo clippy`)
- [x] Code is formatted (`cargo fmt`)
- [x] All relevant tests pass
- [x] Self-reviewed my code
- [x] Comments added where the no-evidence fallback is non-obvious
- [x] No hardcoded secrets or sensitive data

### DIFF
--- a/crates/dataprof-partial/src/lib.rs
+++ b/crates/dataprof-partial/src/lib.rs
@@ -1039,13 +1039,17 @@ fn schema_from_streaming_stats(
 ) -> SchemaResult {
     let columns = headers
         .iter()
-        .filter_map(|name| {
-            column_stats
-                .get_column_stats(name)
-                .map(|stats| ColumnSchema {
-                    name: name.clone(),
-                    data_type: profile_builder::infer_data_type_streaming(stats),
-                })
+        .map(|name| {
+            // A header can declare a column before any row creates statistics.
+            // String is the existing type convention when no values were observed.
+            let data_type = column_stats.get_column_stats(name).map_or(
+                dataprof_core::DataType::String,
+                profile_builder::infer_data_type_streaming,
+            );
+            ColumnSchema {
+                name: name.clone(),
+                data_type,
+            }
         })
         .collect();
 
@@ -1252,6 +1256,51 @@ mod tests {
         assert_eq!(result.columns[2].name, "salary");
         assert_eq!(result.columns[2].data_type, DataType::Float);
         assert!(result.rows_sampled > 0);
+    }
+
+    #[test]
+    fn test_infer_schema_header_only_csv_preserves_declared_columns() {
+        let f = write_temp_csv("name,age\n");
+
+        let schema = infer_schema(f.path()).unwrap();
+        let structure = analyze_structure(f.path(), None).unwrap();
+
+        assert_eq!(schema.rows_sampled, 0);
+        assert!(schema.schema_stable);
+        assert_eq!(
+            schema
+                .columns
+                .iter()
+                .map(|column| (column.name.as_str(), column.data_type.clone()))
+                .collect::<Vec<_>>(),
+            vec![("name", DataType::String), ("age", DataType::String),]
+        );
+        assert_eq!(
+            schema
+                .columns
+                .iter()
+                .map(|column| column.name.as_str())
+                .collect::<Vec<_>>(),
+            structure
+                .columns
+                .iter()
+                .map(|column| column.name.as_str())
+                .collect::<Vec<_>>()
+        );
+    }
+
+    #[test]
+    fn test_infer_schema_empty_json_formats_have_no_columns() {
+        for (label, file) in [
+            ("json", write_temp_json("[]")),
+            ("jsonl", write_temp_jsonl("")),
+        ] {
+            let schema = infer_schema(file.path()).unwrap();
+
+            assert_eq!(schema.rows_sampled, 0, "{label}");
+            assert!(schema.schema_stable, "{label}");
+            assert!(schema.columns.is_empty(), "{label}");
+        }
     }
 
     #[test]
@@ -1884,6 +1933,23 @@ mod async_tests {
         assert_eq!(result.columns[2].data_type, DataType::Float);
         assert!(result.rows_sampled > 0);
         assert!(result.schema_stable); // small data — all rows consumed
+    }
+
+    #[tokio::test]
+    async fn test_infer_schema_stream_header_only_csv_preserves_declared_columns() {
+        let source = csv_source(b"name,age\n");
+        let result = infer_schema_stream(source).await.unwrap();
+
+        assert_eq!(result.rows_sampled, 0);
+        assert!(result.schema_stable);
+        assert_eq!(
+            result
+                .columns
+                .iter()
+                .map(|column| (column.name.as_str(), column.data_type.clone()))
+                .collect::<Vec<_>>(),
+            vec![("name", DataType::String), ("age", DataType::String),]
+        );
     }
 
     #[tokio::test]
--- a/python/tests/test_python_api.py
+++ b/python/tests/test_python_api.py
@@ -1533,6 +1533,21 @@ class TestPartialAnalysis:
         assert len(result.column_names) == result.num_columns
         assert result.rows_sampled > 0
 
+    def test_header_only_csv_surfaces_agree_on_declared_schema(self, tmp_path):
+        path = tmp_path / "header_only.csv"
+        path.write_text("name,age\n")
+
+        profile = dataprof.profile(path)
+        schema = dataprof.infer_schema(path)
+        structure = dataprof.analyze_structure(path)
+
+        expected_names = ["name", "age"]
+        assert profile.rows == schema.rows_sampled == structure.rows_sampled == 0
+        assert list(profile) == list(schema.column_names) == expected_names
+        assert [column.name for column in structure.columns] == expected_names
+        assert [column["data_type"] for column in schema.columns] == ["string", "string"]
+        assert schema.schema_stable is True
+
     def test_infer_schema_path_object(self):
         result = dataprof.infer_schema(Path(CSV_FILE))
         assert result.num_columns > 0
--- a/python/tests/test_python_api.py
+++ b/python/tests/test_python_api.py
@@ -1546,6 +1546,8 @@ class TestPartialAnalysis:
         assert list(profile) == list(schema.column_names) == expected_names
         assert [column.name for column in structure.columns] == expected_names
         assert [column["data_type"] for column in schema.columns] == ["string", "string"]
+        assert [profile[name].data_type for name in expected_names] == ["string", "string"]
+        assert [column.data_type for column in structure.columns] == ["string", "string"]
         assert schema.schema_stable is True
 
     def test_infer_schema_path_object(self):


crates/dataprof-partial/src/lib.rs
```diff
@@ -1039,13 +1039,17 @@ fn schema_from_streaming_stats(
...
     let columns = headers
         .iter()
-        .filter_map(|name| {
-            column_stats
-                .get_column_stats(name)
-                .map(|stats| ColumnSchema {
-                    name: name.clone(),
-                    data_type: profile_builder::infer_data_type_streaming(stats),
-                })
+        .map(|name| {
```
AndreaBozzo, line 1042: This fixes `infer_schema`, but the layer underneath still drops the columns, and it is the layer everything shares. `analyze_csv_from_reader` builds its `StreamingColumnCollection` without pre-registering the headers it just read and validated, so `dataprof::analyze_csv_file` (publicly re-exported at `crates/dataprof/src/lib.rs:35`) still reports zero columns for a header-only CSV while every engine reports two. Measured on this branch:

```
analyze_csv_file:  cols=[]                quality=None
Incremental:       cols=["name", "age"]   quality=Some((0.0, "Exact"))
Columnar:          cols=["name", "age"]   quality=Some((0.0, "Exact"))
```

The engines get it right because `incremental.rs:168` already calls `column_stats.init_columns(&names)` right after `validate_unique_column_names`. Doing the same in `analyze_csv_from_reader_with_hints` fixes both cases at once:

```rust
    let mut column_stats = StreamingColumnCollection::new().with_semantic_hints(semantic_hints);
    // The header declares the columns; a file with zero data rows still has them.
    column_stats.init_columns(&header_names);
    let mut rows_read = 0;
```

I tried it: all three tests added in this PR pass with the `schema_from_streaming_stats` change reverted to `filter_map`, `analyze_csv_file` comes into line with the engines, and the whole workspace is green except `crates/dataprof-csv/src/lib.rs:567` (`test_analyze_csv_file_empty_file`), which asserts `column_profiles.len() == 0` for `"name,age
"` — i.e. the one test that pins the inconsistency. It also makes two existing special cases redundant: `analyze_structure_csv`'s `if columns.is_empty() && !headers.is_empty()` (line 809) and `empty_structure_columns_from_headers` (line 954) could be deleted, and the "no observations -> String" rule would live only in `infer_data_type_streaming` instead of the three places it now appears.

That is a second behavior change (`analyze_csv_file` output), so a follow-up PR is a fair call under the one-behavior-change rule. But then the summary here should not say "preserve declared CSV columns when a file or stream contains headers but no data rows" without qualification — the CSV parser entry point still does not.

python/tests/test_python_api.py
```diff
@@ -1533,6 +1533,21 @@ def test_infer_schema(self):
...
+
+        profile = dataprof.profile(path)
+        schema = dataprof.infer_schema(path)
+        structure = dataprof.analyze_structure(path)
+
+        expected_names = ["name", "age"]
+        assert profile.rows == schema.rows_sampled == structure.rows_sampled == 0
+        assert list(profile) == list(schema.column_names) == expected_names
+        assert [column.name for column in structure.columns] == expected_names
+        assert [column["data_type"] for column in schema.columns] == ["string", "string"]
```
AndreaBozzo, line 1548: The test is named for agreement across surfaces but only asserts types on one of them. `profile` and `analyze_structure` both report `string` here (I checked), so the parity claim can be made exact:

```python
        assert [column["data_type"] for column in schema.columns] == ["string", "string"]
        assert [profile[name].data_type for name in expected_names] == ["string", "string"]
        assert [column.data_type for column in structure.columns] == ["string", "string"]
```

Without this, a regression that changed only the `profile` fallback type would still pass.

AndreaBozzo (commented):
Copilot's pass returned no comments; here is a manual review on top of it.

**The fix itself is correct and the tests discriminate.** I verified the fallback is only reachable at zero rows: `StreamingColumnCollection::process_record` creates an entry for every header on the first record, so with >=1 data row no header can be missing from `column_stats`. Reverting only the production hunk and keeping the tests makes both header-only tests fail (`left: []`, `right: [("name", String), ("age", String)]`); the empty-JSON test passes either way, which is the right shape for a guard against over-correcting. `DataType::String` is genuinely the existing no-evidence convention (`infer_data_type_streaming` falls through to it, and `empty_structure_columns_from_headers` already hardcodes it).

Commands run: `cargo test -p dataprof-partial --all-features` (44 + 4 doc, green), `cargo test --workspace --all-features`, `uv run maturin develop`, `uv run pytest python/tests/test_python_api.py -k "header_only or TestPartialAnalysis"` (11 passed).

**One substantive gap** (inline below): the CSV parser path that every surface shares still drops the columns, so `dataprof::analyze_csv_file` disagrees with `Profiler` and with the three surfaces this PR just aligned. That is the same output-contract rule from AGENTS.md, one layer down.

Also worth a separate issue (pre-existing, not caused here): a header-only CSV with two or more trailing blank lines resolves to the wrong delimiter, so the schema becomes a single column literally named `name,age`. `detect_delimiter` scores the blank records in `split_sample_records` output, and for `"name,age


"` the modal field count for `,` is 1 (two blanks vs one header), which loses to `;` scoring 1 field consistently across all three records. Before this PR `infer_schema` returned zero columns there, so the change makes a pre-existing bug visible rather than introducing one, and surfacing beats silence. One-line fix, all 40 `dataprof-csv` tests stay green:

```rust
// crates/dataprof-csv/src/lib.rs, in detect_delimiter
for record in records.iter().filter(|r| !r.is_empty()).take(5) {
```

Verified after that change: `"name,age


"` -> `,` / `["name", "age"]`, and `;`/tab detection is unaffected.

AndreaBozzo:
Filed the follow-up as #558 (milestoned 0.11.0), covering the `analyze_csv_from_reader` gap from the review above. It carries the measured before/after, the `init_columns` fix, the `dataprof-csv/src/lib.rs:567` assertion that needs updating, and the two `dataprof-partial` helpers that become deletable. Nothing here blocks this PR.

crates/dataprof-partial/src/lib.rs
floze-the-genius: Agreed. I narrowed the PR summary and notes to partial schema inference and linked #558 for the direct analyze_csv_file parser path, so this PR no longer claims that entry point.

python/tests/test_python_api.py
floze-the-genius: Added the exact profile and analyze_structure type assertions in 37bd505. The focused parity test and Ruff checks pass.

AndreaBozzo (approved):
Approving. Beyond the green checks I reverted the `filter_map` → `map` change while keeping the new tests, and both header-only tests fail as they should (42 passed, 2 failed: the sync `test_infer_schema_header_only_csv_preserves_declared_columns` and the async `test_infer_schema_stream_header_only_csv_preserves_declared_columns`). `test_infer_schema_empty_json_formats_have_no_columns` passes either way, which is correct — it pins behaviour that must be preserved, and the description presents it that way rather than claiming it as coverage of the fix.

Blast radius checks out as confined to CSV: `infer_schema_from_json_reader` passes `column_stats.column_names()` as its `headers`, so every name there is guaranteed a stats entry and the new fallback is structurally unreachable on that path.

One item for the follow-up rather than for this PR. The `map_or` fallback becomes dead code the moment #558 lands: once `analyze_csv_from_reader` calls `init_columns(&header_names)`, every header has a stats entry, `get_column_stats` never returns `None`, and the `DataType::String` default cannot be reached. I checked the two agree — an empty `StreamingStatistics` has no non-empty sample values, so `infer_data_type_streaming` falls through to `String`, the same answer this fallback hard-codes. That means the tests added here survive #558 unchanged and only the fallback itself needs removing. I will add `schema_from_streaming_stats` to #558's deletion list, which already tracks collapsing the "no observations means `String`" rule back to a single definition.

Nothing else to change. I had noted the `dataprof_core::DataType::String` qualification as redundant given the import at the top of the file, but on a closer look the one other place in this file that spells the same no-evidence fallback — `empty_structure_columns_from_headers` — uses exactly that form, so this matches its neighbour and the bare spelling only appears in the test modules, which import `DataType` directly. Leaving it consistent with the surrounding code is the better outcome, and #558 deletes both spellings anyway.

Thanks for the clean scoping — keeping the parser entry point out of this and letting it be tracked separately was the right call.

PR #575, opened 2026-08-12
fix(async): reject duplicate CSV headers

### LINKED ISSUE #574: The async streaming reader accepts duplicate CSV headers and merges two columns into one profile
## Summary

The async streaming reader never validates that CSV headers are unique. Every
other CSV path does. A stream whose header repeats a name silently merges two
source columns into one profile with doubled counts, and the report then claims
fewer columns than the source has.

## Mechanism

`validate_unique_column_names` is called on the sync parser
(`crates/dataprof-csv/src/lib.rs`) and on the incremental engine
(`crates/dataprof-engines/src/streaming/incremental.rs`), in both cases
immediately before `column_stats.init_columns(&names)`, with a comment saying
why the order matters:

```rust
// Reject duplicate headers before any per-column state is built:
// init_columns keys on name, so a repeat would otherwise merge
// two source columns into one profile with an inflated count.
dataprof_core::validate_unique_column_names(&names, "CSV header")?;
column_stats.init_columns(&names);
```

`AsyncStreamingProfiler::process_chunks`
(`crates/dataprof-engines/src/streaming/async_reader.rs`) has the `init_columns`
call but not the guard. `StreamingColumnCollection` keys on name in both
`init_columns` and `process_record`, so the duplicate collapses either way; the
missing validation is what lets it through silently.

This predates #572, which only added the `init_columns` call to the async path
and does not change what happens with a duplicate header.

## Why it matters

A profiler reporting one column where the source declared two, with counts that
are the sum of both, is a wrong answer that looks entirely plausible. The sync
paths already decided this is an error worth refusing; the async path should
give the same answer to the same bytes.

## Definition of done

- A CSV stream with a repeated header name fails on the async path with the
  same error the sync parser raises for the same bytes.
- A regression test covering it, next to the existing sync coverage in
  `crates/dataprof-core/src/validation.rs`.

aryansk:
## Description

Reject duplicate CSV header names in the async streaming reader before per-column state is initialized. This keeps async CSV behavior aligned with the existing file, incremental, and columnar paths and prevents two source columns from being merged into one profile.

## Related issue

Fixes #574

## Testing

- `cargo test --features async-streaming --test async_streaming test_profiler_profile_stream_rejects_duplicate_csv_headers`
- `cargo fmt --all`

Prepared with AI assistance; maintainer review is requested.

### DIFF
--- a/crates/dataprof-engines/src/streaming/async_reader.rs
+++ b/crates/dataprof-engines/src/streaming/async_reader.rs
@@ -576,6 +576,7 @@ impl AsyncStreamingProfiler {
             .map_err(|e| Self::csv_error(&e, flexible))?;
 
         let header_fields: Vec<String> = headers.iter().map(|f| f.to_string()).collect();
+        dataprof_core::validate_unique_column_names(&header_fields, "CSV header")?;
         let header_len = header_fields.len();
         // Byte counts come from the parser's own position rather than from the
         // parsed fields: a record's fields exclude delimiters, quotes and line
--- a/tests/async_streaming.rs
+++ b/tests/async_streaming.rs
@@ -213,6 +213,23 @@ async fn test_profiler_profile_stream_header_only_csv_preserves_schema() {
     assert_eq!(report.execution.rows_processed, 0);
 }
 
+#[tokio::test]
+async fn test_profiler_profile_stream_rejects_duplicate_csv_headers() {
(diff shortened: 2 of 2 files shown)


AndreaBozzo (approved):
Reviewed: matches the None-means-unassessed contract, cross-engine test asserts raw and serialized presence. CI green (23/23).

PR #576, opened 2026-08-12
fix(metrics): align empty CSV quality parity

### LINKED ISSUE #573: analyze_csv_file returns no quality block for an empty CSV; every engine returns one scoring 0.0
## Summary

`analyze_csv_file` returns `quality: None` for a zero-byte CSV. Every engine
returns a full quality block for the same file, scoring `0.0` with confidence
`Exact` over zero columns and zero rows. This is the residual half of #558:
that issue aligned the header-only case across all paths, and the empty-file
case was left as it was.

## Measured on the #558 branch (all paths otherwise aligned)

```
header-only  standard:                    cols=2 quality=Some((0.0, Exact))
header-only  Auto/Incremental/Columnar:   cols=2 quality=Some((0.0, Exact))

empty        standard:                    cols=0 quality=None
empty        Auto/Incremental/Columnar:   cols=0 quality=Some((0.0, Exact))
```

## Mechanism

`crates/dataprof-csv/src/lib.rs` short-circuits before the assembler when no
column was declared:

```rust
if column_profiles.is_empty() && rows_read == 0 {
    return Ok(ReportAssembler::new(...).skip_quality().build());
}
```

The engines have no such early return. They only call `skip_quality()` when the
caller excluded the quality metric pack, so an empty source flows through the
assembler and aggregates seven unassessed dimensions to `0.0`.

## Why it matters

`None` means "not analyzed" and the parser path is right: nothing was assessed.
The engines assert a measurement of the worst possible quality about a file with
no data in it, and that value is what a serialized report carries.

## Relation to other issues

- #571 changes the *score* of an empty aggregate from `0.0` to `None`. It does
  not close this: the engines would still emit a `quality` block where the
  parser emits none, so the two surfaces would still disagree about whether the
  file was assessed at all. Deciding this alongside #571 is reasonable, since
  the honest answer for both is probably the same one.
- Split out of the review of #572, which fixes the header-only half.
  `test_analyze_csv_file_empty_file_has_no_columns` there pins the parser side
  and carries a comment pointing here, so whichever way this is decided, the
  test moves with it.

## Definition of done

- A zero-byte CSV produces the same `quality` presence and score on
  `analyze_csv_file` and on all three engines, raw and serialized.
- A test asserts that parity rather than asserting one path's answer.

aryansk:
## Description

Treat a zero-byte CSV as unassessed across standard, incremental, and columnar analysis paths, matching the parser’s `quality: None` contract. Added raw and serialized cross-engine parity coverage.

## Related issue

Fixes #573

## Testing

- `cargo test --test cross_engine_consistency test_empty_csv_quality_presence_matches_across_engines_and_serialization`
- `cargo fmt --all`

Prepared with AI assistance; maintainer review is requested.

### DIFF
--- a/crates/dataprof-engines/src/streaming/incremental.rs
+++ b/crates/dataprof-engines/src/streaming/incremental.rs
@@ -397,6 +397,7 @@ impl IncrementalProfiler {
             execution = execution.with_source_exhausted(false);
         }
 
+        let empty_source = column_profiles.is_empty() && analyzed_rows == 0;
         let mut assembler = ReportAssembler::new(
             DataSource::File {
                 path: file_path.display().to_string(),
@@ -409,7 +410,7 @@ impl IncrementalProfiler {
         )
         .columns(column_profiles);
 
-        if MetricPack::include_quality(packs) {
+        if MetricPack::include_quality(packs) && !empty_source {
             assembler = assembler
                 .with_quality_data(sample_columns)
                 .with_row_duplicates(column_stats.row_duplicate_summary())
--- a/crates/dataprof-parquet/src/arrow_profiler.rs
+++ b/crates/dataprof-parquet/src/arrow_profiler.rs
@@ -322,6 +322,7 @@ impl ArrowProfiler {
             execution = execution.with_truncation(TruncationReason::MaxRows(max as u64));
         }
 
+        let empty_source = column_profiles.is_empty() && total_rows == 0;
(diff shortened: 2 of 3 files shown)


AndreaBozzo (approved):
Reviewed: matches the None-means-unassessed contract, cross-engine test asserts raw and serialized presence. CI green (23/23).

PR #578, opened 2026-08-12
feat(profile): flag a column that has no dominant type

### LINKED ISSUE #544: Quality score is non-monotonic: 20% junk scores 100.0, 18% junk scores 95.5
## Summary

The overall quality score is not monotonic in data quality. A column with 20%
non-numeric junk scores **100.0**, while the same column with 18% junk scores
**95.5**. A column that is 50% junk also scores **100.0**.

Adding bad data to a dataset can raise its quality score, and can raise it to a
perfect one. For a tool whose headline output is a data-quality number, this
inverts the meaning of the number precisely in the range where a user most needs
it to be right.

## Reproduction

```python
import dataprof as dp
import io

for junk in (0, 100, 180, 190, 200, 250, 500):
    vals = [str(1000 + i) for i in range(1000 - junk)] + [f"junk{i}" for i in range(junk)]
    csv = ("v\n" + "\n".join(vals) + "\n").encode()
    r = dp.profile(io.BytesIO(csv), format="csv")
    print(f"{junk/10:5.1f}% junk  type={[p.data_type for p in r.profiles][0]:<8}"
          f" data_type_consistency={r.quality.consistency['data_type_consistency']:>6}"
          f" overall={r.quality_score}")
```

```
  0.0% junk  type=integer  data_type_consistency= 100.0 overall=100.0
 10.0% junk  type=float    data_type_consistency=  90.0 overall=97.5
 18.0% junk  type=float    data_type_consistency=  82.0 overall=95.5
 19.0% junk  type=float    data_type_consistency=  81.0 overall=95.25
 20.0% junk  type=string   data_type_consistency= 100.0 overall=100.0
 25.0% junk  type=string   data_type_consistency= 100.0 overall=100.0
 50.0% junk  type=string   data_type_consistency= 100.0 overall=100.0
```

The score decreases monotonically to 95.25 at 19% junk, then jumps to a perfect
100.0 at 20% and stays there no matter how much worse the data gets.

## Mechanism

`infer_type` in `crates/dataprof-metrics/src/analysis/inference.rs:67` classifies
a column as `Float` only while more than 80% of non-null values parse as numeric:

```rust
// 80% threshold: tolerates a few non-numeric values (e.g. "N/A", missing)
if float_count as f64 / non_empty.len() as f64 > 0.8 {
    return DataType::Float;
}
```

Below that ratio the column falls back to `DataType::String`. `data_type_consistency`
then measures conformance to the inferred type, and every value conforms to
`String` by construction, so consistency reports 100%. The consistency dimension
feeds the overall score, which becomes perfect.

The metric is measured against a type that is itself chosen by majority vote of
the same values, so it cannot report the mess it exists to detect. Crossing the
threshold does not merely lose precision, it erases the finding.

## Agent-facing impact

`to_llm_context()` passes the inverted number straight to a model, with no flag
on the column:

```
--- 18% junk ---
rows: 1,000 | columns: 1 | quality: 95.5/100
schema (1):
- v: float

--- 50% junk ---
rows: 1,000 | columns: 1 | quality: 100.0/100
schema (1):
- v: string
```

An agent asking dataprof whether a half-garbage column is safe to use is told the
data is perfect.

## Why this is a product decision, not just a patch

The fix is not simply lowering or raising the 0.8 threshold. Any threshold has
the same cliff, only in a different place. The question to settle is what
consistency means for a column that has no dominant type:

1. Record the runner-up type share and report `data_type_consistency` against the
   dominant *parseable* type even when the column falls back to `String`, so a
   50/50 column reports roughly 50% rather than 100%.
2. Introduce an explicit `Mixed` type distinct from `String`, so a column that
   defeated inference is visibly different from a column that is genuinely text,
   and score it accordingly.
3. Keep the fallback but exclude `String` columns that lost a numeric majority
   from the consistency dimension entirely (`None`, "not analyzed"), which is
   honest under the repo's `None`-versus-empty convention and stops the perfect
   score, though it also hides the problem.

Option 1 or 2 preserves the invariant worth having: **adding bad values to a
column must never increase its quality score.** That invariant is testable
directly and would make a good regression property test independent of whichever
option is chosen.

- A monotonicity property test over junk ratio from 0% to 100%: the overall score
  and the consistency dimension must be non-increasing.
- A genuinely textual column (free text, no numeric majority ever) still scores
  as consistent, so the fix does not penalise ordinary string columns.
- Cross-engine parity holds for whatever the new consistency value is.
- `to_llm_context()` flags a column that defeated type inference.

Found during 0.10.0 dogfooding.

### LINKED ISSUE #561: to_llm_context() does not flag a column that defeated type inference
## Summary

`to_llm_context()` gives an agent the dataset quality score and each column's
type, but nothing that says a column had no dominant type. After #560 the score
is no longer inverted, so a half-numeric column reads 86.7/100 rather than
100/100, but the column itself still renders as an ordinary `string`:

```
rows: 1,000 | columns: 1 | quality: 86.7/100

schema (1):
- v: string
```

An agent deciding whether `v` is safe to use cannot tell that column apart from
a column of names. This is acceptance criterion 4 of #544, split out because it
needs a new per-column field rather than a scoring change.

## Why it needs new plumbing

Nothing on `ColumnProfile` carries the evidence today:

- `data_type` is `String` for both a genuinely textual column and one that
  defeated inference.
- `invalid_count` is `None` for string columns by contract — it is only computed
  for numeric and date columns, and `None` means "the check did not run".

So the dominant lexical class and its share, which `dataprof-metrics` already
computes per column inside `calculate_type_consistency`, are discarded after
being folded into the dataset-level `data_type_consistency` average.

## Sketch

Surface the per-column evidence the consistency calculator already has, as an
additive optional field on `ColumnProfile` — `None` keeps meaning "not
analyzed", so nothing existing has to change shape:

```rust
/// Share of non-null values in the column's largest lexical class, and which
/// class that is. `None` when type consistency was not assessed.
pub type_homogeneity: Option<TypeHomogeneity>,
```

Then `to_llm_context()` can flag a column whose dominant class holds less than
all of its values. Open questions for whoever picks this up:

- Does the flag belong on every mixed column, or only below some share? A
  threshold here is display-only and does not affect the score, so unlike #544
  it is a legitimate place for one.
- Wording: the flag should say what is actually true ("60% numeric, 40% text"),
  not restate the type.
- The value is computed over reservoir samples (10k), so the flag inherits
  sample scope and should disclose it the way other sampled fields do.

## Acceptance criteria

- A column with no dominant type is visibly distinct from a textual column in
  `to_llm_context()`.
- The new field round-trips through `to_json()` / report reload, and `None`
  still means "not analyzed".
- Cross-engine parity: every engine reports the same value for the same input.
- The `dataprof` skill's #544 warning (added in #555) is updated or removed once
  the flag makes the caveat unnecessary.

Split out of #544; the scoring half of that issue is #560.

AndreaBozzo:
## Summary

`to_llm_context()` gave an agent the dataset quality score and each column's
type, and nothing that said a column had no dominant form. After #560 a
half-numeric column scores 86.7 rather than 100, but the column still rendered
as `- v: string` — indistinguishable from a column of names.

`data_type` cannot carry that difference, and `invalid_count` is `None` on
string columns by contract, so the dominant lexical class the consistency
calculator already computed was folded into a dataset average and discarded.
This surfaces it per column and flags it.

## Before / after

`.claude/skills/dataprof/evals/fixtures/payments_mixed_amount.csv` — a money
column where 4 of 10 values are placeholders:

```
 rows: 10 | columns: 4 | quality: 97.9/100
                                                 flags (1):
                                               + - amount_eur: mixed types (60% numeric, 40% text)

 schema (4):
 - amount_eur: string
```

## What changed

- **`ColumnProfile.type_homogeneity`** — non-null values counted by lexical
  class, `{"numeric": 6, "date": 0, "boolean": 0, "text": 4}`. Additive and
  optional: `None` keeps meaning "not analyzed", while a column that *was*
  classified and had nothing to classify is `Some` with four zero counts. Counts
  rather than a share, so the flag can state the mixture it actually found
  instead of a dominant class and an unnamed remainder.
- **One classifier.** `lexical_class` / `classify_lexical_forms` move out of the
  consistency calculator into `dataprof-metrics::analysis::inference`, so a
  column's flag and its consistency score are the same measurement rather than
  two that agree by inspection.
- **Every engine.** Both profile construction paths populate the field — the
  streaming builder from its retained sample, the in-memory path (database
  connectors) from the whole column. Not gated by the metric packs: which forms
  a column holds is schema-level evidence like the inferred type, not a
  statistic.
- **The flag.** Emitted once at least 5% of classified values fall outside the
  dominant class. Display only — it changes no score — and `identifier` columns
  are exempt, matching the exemption consistency already makes for ID schemes
  that mix forms on purpose.
- **Sample scope.** The counts cover what the profiler retained, so the flag
  appends `sampled 10,000 of 50,000 values` when it was short of the column, and
  summing the counts against `total_count - null_count` says so directly.
- **Flat exports** (`to_dataframe`, `save(".csv")`, `to_arrow`) cannot hold the
  count map, so they carry `dominant_type` and `dominant_type_share`.

Two open questions from the issue, settled: the flag threshold (5% outside the
dominant class, display only) and the field shape (per-class counts rather than
dominant + share).

## Acceptance criteria

- [x] A column with no dominant type is visibly distinct from a textual column
      in `to_llm_context()`.
- [x] The field round-trips through `to_json()` / report reload, and `None`
      still means "not analyzed" — covered by the generic round-trip parity
      sweep plus explicit absence-vs-zero tests.
- [x] Cross-engine parity: `type_homogeneity_agrees_across_engines` profiles one
      mixed CSV through standard / auto / incremental / columnar and asserts
      identical counts, including through serialization. Verified by hand across
      JSON, JSONL, Parquet and pandas too.
- [x] The `dataprof` skill's #544 warning is updated: SKILL.md and
      `reference/interpretation.md` now teach the field and the flag, and the
      `high-score-is-not-clean` eval no longer says the flag is missing.

## Tests

Each part was reverted separately and the guards confirmed failing:

| Reverted | Fails |
| --- | --- |
| runtime classification (`build_column_profile`) | `type_homogeneity_agrees_across_engines`, 12 of 17 Python tests |
| in-memory classification (`analyze_column`) | 3 `analysis::column` tests |
| the flag in `_column_flags` | 7 Python flag tests |
| threshold 5% → 20% | 5 Python flag tests |
| `column_to_dict` emission | 11 round-trip tests |
| read-back normalization | the malformed-mapping test |

`test_eval_rubrics_cite_current_fixture_values` now also profiles the fixture
and checks the flag wording the rubric quotes, so that rubric cannot go stale
the way the score did under #544.

## Commands run

```
cargo fmt --all
cargo test --workspace
cargo run --example generate_profile_schema     # committed schema regenerated
uv run maturin develop
uv run pytest python/tests/ -q                  # 853 passed
uv run ty check python/
```

Closes #561. Closes #544 — this was its remaining acceptance criterion.

### DIFF
--- a/.claude/skills/dataprof/SKILL.md
+++ b/.claude/skills/dataprof/SKILL.md
@@ -150,9 +150,22 @@ re-read and eyeball two files by hand.
   A perfect score therefore means the column holds one kind of value — for a
   `string` column, ordinary text — not that the text is fit for your purpose. And
   the score is symmetric around a 50/50 mix: 20% junk and 80% junk both report
-  80. If a column is typed `string` but its name or data suggests it should hold
-  numbers, dates, or identifiers, say so; read `sample_values` rather than the
-  score alone.
+  80.
+- **`data_type` alone does not tell you whether a column is uniform.** A column
+  of names and a column that is 60% numbers are both `string`. The per-column
+  evidence is `type_homogeneity` — counts of non-null values by lexical class —
+  and `to_llm_context()` renders it as a `mixed types` flag once at least 5% of
+  the values fall outside the dominant class:
+
+  ```
+  flags (1):
+  - amount_eur: mixed types (60% numeric, 40% text)
+  ```
+
+  Report that mixture rather than the type. The counts come from the values the
+  profiler retained, so on a large source they describe the 10k reservoir
+  sample; the flag says `sampled N of M values` when that is the case, and
+  summing the counts against `total_count - null_count` tells you directly.
 
 Before interpreting a specific dimension, an approximate count, a detected
 pattern, or a comparison, read [reference/interpretation.md](reference/interpretation.md).
--- a/.claude/skills/dataprof/evals/README.md
+++ b/.claude/skills/dataprof/evals/README.md
@@ -51,9 +51,9 @@ score improvement has missed the row loss.
 the column `string`. Consistency now scores that column on its dominant lexical
 class, so it contributes 6 of 10 values and the file scores **97.9/100** rather
 than the 100.0/100 it reported before #544 was fixed. That is still high enough
-to read as clean, and `to_llm_context()` still shows no per-column flag (#561), so
-a money column carrying text remains a finding the score alone will not make for
-you.
+to read as clean, which is the point of the scenario: the finding is in
+`to_llm_context()`'s per-column flag —
+`amount_eur: mixed types (60% numeric, 40% text)` — not in the score.
 
 ## When these change
 
--- a/.claude/skills/dataprof/evals/scenarios.json
+++ b/.claude/skills/dataprof/evals/scenarios.json
@@ -59,6 +59,7 @@
     "expected_behavior": [
       "Profiles the file rather than eyeballing ten rows",
       "Notices that amount_eur is typed string, not float, and that a money column typed as text is the finding",
+      "Surfaces the mixed types flag on amount_eur (60% numeric, 40% text) rather than reading the schema line alone",
       "Names the four unparseable values (pending, n/a, see invoice, awaiting PO) as the reason no numeric statistics exist for that column",
       "Explains that the 97.9/100 score does not contradict this: the column is scored on its dominant lexical class, so 6 numeric values out of 10 leave the file looking mostly fine",
       "Answers the actual question with a no, or a qualified no, for a finance load"
--- a/.claude/skills/dataprof/reference/interpretation.md
+++ b/.claude/skills/dataprof/reference/interpretation.md
@@ -8,6 +8,7 @@ means and — more importantly — what it does not.
 
 - The absence rule
 - Quality dimensions, one by one
+- Type homogeneity
 - Approximation provenance
 - Patterns and sensitive data
 - Comparisons
@@ -61,9 +62,11 @@ perfect score means one kind — which for a `string` column is ordinary text, n
 a guarantee the text is useful.
 
 The score is symmetric around a 50/50 mix: 20% junk and 80% junk both report 80,
-because the minority is what costs the column its consistency in each case. Read
-it with the column's `data_type` and its `sample_values`, not alone. A column
-that is 80% junk is typed `string` because that is what it mostly is.
+because the minority is what costs the column its consistency in each case. A
+column that is 80% junk is typed `string` because that is what it mostly is.
+
+The dataset-level score cannot tell you *which* column is mixed; the per-column
+evidence is `type_homogeneity` (see below). Read the two together.
 
 Two kinds of column keep the plain type check rather than the dominant-class
 rule, so a low score there means something different: columns declared through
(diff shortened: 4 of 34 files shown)


AndreaBozzo:
Both Copilot findings applied in 3e09e29.

**Python dialect schema shape** — real hole, and worse than cosmetic: `{"numeric": 3}` validated against the published schema while `_homogeneity_counts` treats a partial mapping as absence, so a document could be schema-valid and still lose the field on reload. Both dialects now share the `TypeHomogeneity` definition, whose four keys are required. `test_the_published_schema_rejects_what_the_loader_discards` pins the two sides together, and fails against the previous loose-map schema.

The loader stays lenient on the way in — an incomplete map still reads as "not classified" rather than being completed with zeros, since zero is a claim about the data. What changed is that such a document is no longer *valid*.

**Facade re-export** — correct, and an oversight: `ColumnStats`, `Pattern`, `Quartiles` and every other `ColumnProfile` field type are re-exported, so a consumer could read `type_homogeneity` but not name it. `TypeHomogeneity` and `LexicalClass` are re-exported now, along with `lexical_class` / `classify_lexical_forms` beside the existing `infer_type` and `is_null_like_token`. `parser_and_metrics_reexports_compile` names all four, so the gap cannot reopen silently.

Re-ran `cargo clippy --all --all-targets -- -D warnings`, `cargo test --workspace`, the schema regeneration, `uv run pytest python/tests/ -q` (854 passed), ruff, and ty.

PR #579, opened 2026-08-12
fix(csv): ignore blank records when scoring delimiter candidates

### LINKED ISSUE #559: detect_delimiter counts blank records: a header-only CSV with trailing blank lines collapses into one column
## Summary

`detect_delimiter` scores blank records when choosing a delimiter. When blank
records outnumber real ones, the true delimiter loses to one that splits every
record into a single field, and the whole header collapses into one column whose
name is the entire header line.

## Reproduction

`header_only.csv` (header plus two blank lines):

```
name,age


```

```python
import dataprof
s = dataprof.infer_schema("header_only.csv")
print([c["name"] for c in s.columns])
```

Observed on `master` at 824b0fb:

```
h_1blank.csv   'name,age

'      delim=','  cols=['name', 'age']     # correct
h_2blank.csv   'name,age


'    delim=';'  cols=['name,age']        # wrong
```

`profile` and `analyze_structure` report the same single `name,age` column, so
every surface is consistently wrong.

## Mechanism

`crates/dataprof-csv/src/lib.rs:175-194`. `split_sample_records` keeps blank
records, and the scoring loop counts them:

```rust
for record in records.iter().take(5) {
    let fields = count_fields(record, delimiter as char);
    *counts.entry(fields).or_insert(0usize) += 1;
}
```

For `"name,age

"` the records are `["name,age", "", ""]`:

| delimiter | field counts | modal fields | consistency | score |
| --------- | ------------ | ------------ | ----------- | ----- |
| `,`       | `{2: 1, 1: 2}` | 1 | 2 | `(false, 2, 1)` |
| `;`       | `{1: 3}`       | 1 | 3 | `(false, 3, 1)` |

`,` loses because its 2-field header is outvoted by two 1-field blanks, dropping
`modal_fields > 1` to false; `;` then wins on consistency. The bug needs blank
records to outnumber non-blank ones, so it does not fire once a file has a
couple of real data rows:

```
data1_2blank.csv  'name,age
1,2


'      delim=','  cols=['name', 'age']
data2_3blank.csv  'name,age
x,1
y,2



'  delim=','  cols=['name', 'age']
```

## Proposed fix

A blank record carries no delimiter evidence, so it should not vote:

```rust
for record in records.iter().filter(|r| !r.is_empty()).take(5) {
```

Verified locally: all 40 `dataprof-csv` tests stay green, and detection is
unchanged for the cases that already worked:

```
h2blank.csv     'name,age


'   delim=','   cols=['name', 'age']   # fixed
h1blank.csv     'name,age

'     delim=','   cols=['name', 'age']
semi.csv        'name;age
x;1
'   delim=';'   cols=['name', 'age']
semi_blank.csv  'name;age


'   delim=';'   cols=['name', 'age']
tabs.csv        'a	b
1	2
'      delim='	'  cols=['a', 'b']
```

Note that `.take(5)` currently means "first 5 records including blanks", so the
filter also makes the sample five records of actual evidence, which is the
intent of the cap.

## Relation to #557 / #552

Pre-existing and independent, but #557 changes how it presents. Before #557,
`infer_schema` returned zero columns for a header-only file, so this was
invisible on that surface; after it, `infer_schema` reports one column named
`name,age`. That is an improvement (surfacing beats silence) and not a reason to
hold #557 — it just means the wrong delimiter is now visible in three places
instead of two.

- A header-only CSV with two or more trailing blank lines detects the correct
  delimiter.
- Delimiter detection for `;`, tab, and `|` files is unchanged, including files
  that contain blank lines.
- A test covers the blank-dominated sample directly, since it fails before the
  fix and passes after.

Found while reviewing #557.

aryansk:
## Summary

`detect_delimiter` scored blank records when choosing a delimiter. A header-only CSV with trailing blank lines could lose the true delimiter to one that splits every record into a single field, collapsing the entire header into one column (gh #559).

## Fix

Blank records now carry no delimiter signal: the scoring loop filters out records that are empty after trimming before taking the first five. The rest of the scoring logic is unchanged.

## Validation

- `cargo test -p dataprof-csv` — 42 passed, including the new regression test
- `cargo clippy -p dataprof-csv --all-targets` — clean
- `cargo fmt --all -- --check` — clean
- `git diff --check` — clean

Closes #559

### DIFF
--- a/crates/dataprof-csv/src/lib.rs
+++ b/crates/dataprof-csv/src/lib.rs
@@ -145,10 +145,10 @@ fn split_sample_records(text: &str) -> (Vec<&str>, bool) {
 ///
 /// The sample is split into logical records with quote handling, so quoted
 /// fields containing embedded newlines or delimiter characters do not skew
-/// detection. Tests comma, semicolon, tab, and pipe over the first 5 records:
-/// a delimiter whose modal field count is greater than one always beats one
-/// that only ever yields single-field records, then higher record agreement
-/// and higher field counts win.
+/// detection. Tests comma, semicolon, tab, and pipe over the first 5 non-blank
+/// records: a delimiter whose modal field count is greater than one always
+/// beats one that only ever yields single-field records, then higher record
+/// agreement and higher field counts win.
 ///
 /// Falls back to comma if no candidate splits the sample.
 pub fn detect_delimiter<R: Read>(reader: R) -> std::io::Result<u8> {
@@ -174,7 +174,14 @@ pub fn detect_delimiter<R: Read>(reader: R) -> std::io::Result<u8> {
 
     for &delimiter in &delimiters {
         let mut counts = std::collections::HashMap::new();
-        for record in records.iter().take(5) {
+        // Blank records carry no delimiter signal: a header followed by
+        // trailing blank lines must not outvote the true delimiter with
+        // single-field "records" (gh #559).
+        let scored = records
+            .iter()
+            .filter(|record| !record.trim().is_empty())
+            .take(5);
+        for record in scored {
             let fields = count_fields(record, delimiter as char);
             *counts.entry(fields).or_insert(0usize) += 1;
         }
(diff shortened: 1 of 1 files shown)


AndreaBozzo (approved):
Reviewed: focused fix, shared-path parity confirmed, regression test covers the boundary. CI green (23/23).

PR #580, opened 2026-08-13
fix(partial): report schema_stable from scan exhaustion, not the raw cap

### LINKED ISSUE #553: schema_stable is false for a file read in full at exactly the 1000-row sample cap
## Summary

`SchemaResult.schema_stable` is computed from whether the sample row count
reached the cap, not from whether the source was exhausted:

```rust
// crates/dataprof-partial/src/lib.rs:1052
// schema_stable is true when we read fewer rows than the sample cap
let schema_stable = rows_sampled < SCHEMA_SAMPLE_ROWS;
```

A file with exactly `SCHEMA_SAMPLE_ROWS` (1000) data rows is read in full and
still reports `schema_stable=False`. `analyze_structure` on the same file
correctly reports it as exhausted and not truncated, so the information needed to
answer correctly is already available.

## Reproduction

```python
import dataprof as dp
from pathlib import Path

for n in (999, 1000, 1001):
    p = Path(f"stable_{n}.csv")
    p.write_bytes(("a,b\n" + "".join(f"{i},x{i}\n" for i in range(n))).encode())
    sc = dp.infer_schema(p)
    st = dp.analyze_structure(p)
    print(f"{n:>5} rows: schema_stable={str(sc.schema_stable):5} sampled={sc.rows_sampled:>4}"
          f" | structure: truncated={str(st.truncated):5} exhausted={st.source_exhausted}")
```

```
  999 rows: schema_stable=True  sampled= 999 | structure: truncated=False exhausted=True
 1000 rows: schema_stable=False sampled=1000 | structure: truncated=False exhausted=True
 1001 rows: schema_stable=False sampled=1000 | structure: truncated=True  exhausted=False
```

The 1000-row row is the bug. The file was fully read, `analyze_structure` says so,
and `infer_schema` reports the schema as not stable.

## The naming question underneath it

Even with the off-by-one fixed, the field means "I saw the whole file". That is
what `source_exhausted` already means on `StructureReport`, and it is not what
"schema stable" suggests. A reader reasonably takes `schema_stable=False` to mean
the schema varied across the rows examined, which the field never reports.

The practical effect is that the flag is `False` for essentially every real
dataset, since real datasets exceed 1000 rows. A signal that is always false
carries no information, and a user who filters on it discards every genuine
result.

Two ways to make it honest:

1. Rename to match its meaning (`fully_sampled`, or reuse `source_exhausted` for
   consistency with `StructureReport`) and fix the boundary to test exhaustion
   rather than count.
2. Keep the name and make it mean what it says: track whether the inferred type
   per column changed while consuming the sample, and report instability when it
   did. That is the more useful signal, and it is orthogonal to whether the sample
   was capped, so both fields would be worth having.

Option 2 is the more valuable one for the stated purpose of the partial-analysis
API (decide cheaply whether a full profile is warranted), but option 1 is
correct and small. Doing 1 now does not block 2 later.

- A file with exactly `SCHEMA_SAMPLE_ROWS` rows reports the same
  fully-read/exhausted answer as `analyze_structure` on the same file.
- The boundary is covered by a test at `cap - 1`, `cap`, and `cap + 1`.
- Whatever the field ends up meaning, `SchemaResult` and `StructureReport` agree
  on the same file and use consistent vocabulary.

Found during 0.11 dogfooding of the partial-analysis surface.

aryansk:
## Summary

`schema_stable` was computed from the raw row count (`rows_sampled < SCHEMA_SAMPLE_ROWS`), so a file with exactly 1000 rows — read in full — reported `schema_stable = false` while `analyze_structure` on the same file correctly reported it exhausted (gh #553).

## Fix

`schema_from_profiles` now takes an explicit `truncated` flag and computes `schema_stable = !truncated`. Both the CSV and JSON/JSONL schema paths probe one record past the cap:

- a source with exactly `SCHEMA_SAMPLE_ROWS` records reads `cap` rows → not truncated → `schema_stable = true`
- a source with more reads `cap + 1` → truncated → `schema_stable = false`

`rows_sampled` is clamped back to the cap, so the public count never exceeds 1000. Parquet's metadata-only path is unchanged (`schema_stable = true`).

## Validation

- `cargo test -p dataprof-partial` — 32 passed, including the new `cap`/`cap + 1` boundary regression test
- `cargo test -p dataprof-csv` — 41 passed
- `cargo test -p dataprof-core` — 142 passed
- `cargo test -p dataprof-json` — 52 passed
- `cargo clippy -p dataprof-partial --all-targets` — clean
- `cargo fmt --all -- --check` and `git diff --check` — clean

Closes #553

### DIFF
--- a/crates/dataprof-partial/src/lib.rs
+++ b/crates/dataprof-partial/src/lib.rs
@@ -257,14 +257,21 @@ fn infer_schema_from_csv_reader<R: Read>(
     reader: R,
     has_header: bool,
 ) -> Result<SchemaResult, DataProfilerError> {
+    // Probe one record past the cap: a source with exactly SCHEMA_SAMPLE_ROWS
+    // records reads `cap` (not truncated), anything larger reads `cap + 1`
+    // (truncated). rows_sampled is clamped back to the cap so the public
+    // number never exceeds it.
     let config = CsvParserConfig::default()
         .has_header(has_header)
-        .max_rows(Some(SCHEMA_SAMPLE_ROWS));
+        .max_rows(Some(SCHEMA_SAMPLE_ROWS + 1));
     let (profiles, _column_stats, rows_read, _headers) =
         dataprof_csv::analyze_csv_from_reader(reader, &config)?;
 
+    let rows_sampled = rows_read.min(SCHEMA_SAMPLE_ROWS);
+    let truncated = rows_read > SCHEMA_SAMPLE_ROWS;
+
     // Zero elapsed: the caller patches the timing.
-    Ok(schema_from_profiles(&profiles, rows_read, 0))
+    Ok(schema_from_profiles(&profiles, rows_sampled, truncated, 0))
 }
 
 /// Infer schema from any JSON/JSONL reader. Reads up to `SCHEMA_SAMPLE_ROWS` rows.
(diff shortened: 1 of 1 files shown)


AndreaBozzo (approved):
Reviewed: focused fix, shared-path parity confirmed, regression test covers the boundary. CI green (23/23).

PR #646, opened 2026-08-31
Fix MySQL TIME columns decoding as duration instead of time of day (#644)

### LINKED ISSUE #644: MySQL TIME columns still profile as null: sqlx decodes them as a duration
## Problem

MySQL `TIME` columns still profile as entirely null. #642 added temporal, decimal, UUID and unsigned arms to `db_column_to_string!`, and MySQL `TIME` was the one temporal type it could not cover.

`chrono::NaiveTime` is in the shared chain and handles PostgreSQL `TIME`, but sqlx decodes MySQL `TIME` as a **duration** rather than a time of day, so the `NaiveTime` arm is type-incompatible and the value falls through to `None`.

## Why MySQL is different

MySQL `TIME` is a signed interval, not a clock reading. Its documented range is `-838:59:59` to `838:59:59`, and that is not a corner case a test has to invent:

```
mysql> CREATE TABLE t_probe (t TIME, big TIME);
mysql> INSERT INTO t_probe VALUES ('10:30:00', '838:59:59');
mysql> SELECT * FROM t_probe;
t         big
10:30:00  838:59:59
```

`838:59:59` has no `NaiveTime` representation, so the decode arm cannot simply be "parse it as a clock time".

## What to do

Add a MySQL-specific arm, the same way `u64` and `BigDecimal` are passed per backend from `mysql.rs` (see `db_column_to_string!` in `crates/dataprof-db/src/connectors/common.rs`). The arm decodes sqlx's duration type and renders it.

The rendering is the decision worth making deliberately:

- `HH:MM:SS` with hours allowed past 24 keeps what MySQL stored and round-trips, but a reader may take it for a clock time.
- A signed form (`-01:30:00`) is needed for negative intervals; dropping the sign would be silent corruption of exactly the kind this area has had before.
- Whatever is chosen, the column profiles as **text**, not as a date. A duration is not a date, so it will not feed Timeliness, and that is correct rather than a shortfall.

## Acceptance

- [ ] A MySQL `TIME` column decodes instead of reporting null.
- [ ] The chosen rendering is documented on the macro alongside the existing notes.
- [ ] Cases in `crates/dataprof-db/tests/typed_column_decoding.rs`: an ordinary time, a value past 24 hours, a negative value, and NULL.
- [ ] Both the batch and streaming call sites in `mysql.rs` pass the new arm. There is a test for that shape already (`the_streaming_path_decodes_the_same_types`) — extend it rather than trusting the two sites to stay in step.
- [ ] The remaining-gaps note in `docs/guides/database-connectors.md` loses its MySQL `TIME` entry.

## Getting started

The test suite, the container setup and the per-backend arm mechanism all landed in #642, so this is an arm, a rendering decision, and four test cases. The tests run against a real MySQL: see the `db-integration-tests` job in `.github/workflows/ci.yml` for the service definition, or run `mysql:8.0` locally and set `MYSQL_TEST_URL`.

AnandkumarMall:
Fixes MySQL TIME columns being incorrectly profiled as null because sqlx decodes them as a duration, not a time of day. Modified db_column_to_string! and streaming macros to accept custom formatting functions for specific types, and supplied a custom formatting function for sqlx::types::chrono::TimeDelta that successfully renders the duration mathematically, extracting negative signs and keeping hours uncapped in a standard HH:MM:SS layout.

### DIFF
--- a/crates/dataprof-db/src/connectors/mysql.rs
+++ b/crates/dataprof-db/src/connectors/mysql.rs
@@ -110,7 +110,23 @@ impl DatabaseConnector for MySqlConnector {
             // integer, and without an arm a BIGINT UNSIGNED falls past every
             // signed arm into `bool` and renders as "true" (#365). DECIMAL
             // follows for the same reason PostgreSQL needs it.
-            process_rows_to_columns!(rows, [u64, ::sqlx::types::BigDecimal])
+            // MySQL TIME is a signed interval, decoded as TimeDelta.
+            process_rows_to_columns!(
+                rows,
+                [u64, ::sqlx::types::BigDecimal],
+                [(
+                    ::sqlx::types::chrono::TimeDelta,
+                    |d: ::sqlx::types::chrono::TimeDelta| {
+                        let secs = d.num_seconds();
+                        let sign = if secs < 0 { "-" } else { "" };
+                        let abs_secs = secs.abs();
+                        let hours = abs_secs / 3600;
+                        let minutes = (abs_secs % 3600) / 60;
+                        let seconds = abs_secs % 60;
+                        format!("{}{:02}:{:02}:{:02}", sign, hours, minutes, seconds)
+                    }
+                )]
+            )
         }
 
         #[cfg(not(feature = "mysql"))]
@@ -141,7 +157,19 @@ impl DatabaseConnector for MySqlConnector {
                 batch_size,
                 total_rows,
                 "MySQL",
-                [u64, ::sqlx::types::BigDecimal]
+                [u64, ::sqlx::types::BigDecimal],
+                [(
+                    ::sqlx::types::chrono::TimeDelta,
+                    |d: ::sqlx::types::chrono::TimeDelta| {
+                        let secs = d.num_seconds();
+                        let sign = if secs < 0 { "-" } else { "" };
+                        let abs_secs = secs.abs();
+                        let hours = abs_secs / 3600;
+                        let minutes = (abs_secs % 3600) / 60;
+                        let seconds = abs_secs % 60;
+                        format!("{}{:02}:{:02}:{:02}", sign, hours, minutes, seconds)
+                    }
+                )]
             )
         }
 
--- a/crates/dataprof-db/src/connectors/mysql.rs
+++ b/crates/dataprof-db/src/connectors/mysql.rs
@@ -110,18 +110,25 @@ impl DatabaseConnector for MySqlConnector {
             // integer, and without an arm a BIGINT UNSIGNED falls past every
             // signed arm into `bool` and renders as "true" (#365). DECIMAL
             // follows for the same reason PostgreSQL needs it.
-            // MySQL TIME is a signed interval, decoded as TimeDelta.
+            // MySQL TIME is a signed interval, decoded as a chrono duration (Duration).
             process_rows_to_columns!(
                 rows,
                 [u64, ::sqlx::types::BigDecimal],
                 [(::chrono::Duration, |d: ::chrono::Duration| {
-                    let secs = d.num_seconds();
-                    let sign = if secs < 0 { "-" } else { "" };
-                    let abs_secs = secs.abs();
-                    let hours = abs_secs / 3600;
-                    let minutes = (abs_secs % 3600) / 60;
-                    let seconds = abs_secs % 60;
-                    format!("{}{:02}:{:02}:{:02}", sign, hours, minutes, seconds)
+                    let nanos = d.num_nanoseconds().unwrap_or(0);
+                    let sign = if nanos < 0 { "-" } else { "" };
+                    let abs_nanos = nanos.unsigned_abs();
+                    let total_secs = abs_nanos / 1_000_000_000;
+                    let frac = abs_nanos % 1_000_000_000;
+                    let hours = total_secs / 3600;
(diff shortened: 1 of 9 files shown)


AndreaBozzo:
Thanks for picking this up, @AnandkumarMall! I’ve just merged #642, which touched the same database decoding code and left this PR with merge conflicts. The MySQL TIME-specific change here is still needed. Could you please update the branch with the latest master and resolve the conflicts? Thanks again!

crates/dataprof-db/src/connectors/mysql.rs
```diff
@@ -110,7 +110,20 @@ impl DatabaseConnector for MySqlConnector {
             // integer, and without an arm a BIGINT UNSIGNED falls past every
             // signed arm into `bool` and renders as "true" (#365). DECIMAL
             // follows for the same reason PostgreSQL needs it.
-            process_rows_to_columns!(rows, [u64, ::sqlx::types::BigDecimal])
+            // MySQL TIME is a signed interval, decoded as TimeDelta.
+            process_rows_to_columns!(
+                rows,
+                [u64, ::sqlx::types::BigDecimal],
+                [(::chrono::Duration, |d: ::chrono::Duration| {
```
AndreaBozzo: Could this use `::sqlx::mysql::types::MySqlTime` in the existing backend-specific type list instead? In the locked sqlx 0.8.6, `MySqlTime` implements `Type<MySql>`, `Decode<MySql>`, and `Display`; its formatter handles the signed 838-hour range and nonzero microseconds. Unless two-digit padding for sub-10 hours is an intentional output contract, both call sites could simply add `MySqlTime` alongside `u64` and `BigDecimal`. That would also remove the new mapper macro surface, both duplicated formatter closures, and the direct `chrono` dependency.

crates/dataprof-db/tests/typed_column_decoding.rs
```diff
@@ -251,13 +251,14 @@ mod mysql {
...
-            big BIGINT UNSIGNED
+            big BIGINT UNSIGNED,
+            t TIME(3)
         )";
 
     const ROWS: &str = "INSERT INTO typed_decoding_my VALUES
             (1, '2024-01-15 10:30:00', '2024-01-15 10:30:00', '2024-01-15',
-             1234.5678, 18446744073709551615),
-            (2, NULL, NULL, NULL, NULL, NULL)";
+             1234.5678, 18446744073709551615, '-838:59:59.123'),
```
AndreaBozzo: This value is outside MySQL `TIME` range: the boundary is `-838:59:59.000000`, so adding `.123` exceeds the maximum magnitude. A strict server rejects the fixture insert; a permissive server may clamp it and silently stop testing fractional preservation. Please use `-00:00:00.123` here: it is in range and also pins the important negative-subsecond case, which the old `num_seconds()` implementation rendered without the minus sign. Add `-838:59:59` separately if boundary coverage is desired.

crates/dataprof-db/src/connectors/mysql.rs
```diff
@@ -110,7 +110,27 @@ impl DatabaseConnector for MySqlConnector {
             // integer, and without an arm a BIGINT UNSIGNED falls past every
             // signed arm into `bool` and renders as "true" (#365). DECIMAL
             // follows for the same reason PostgreSQL needs it.
-            process_rows_to_columns!(rows, [u64, ::sqlx::types::BigDecimal])
+            // MySQL TIME is a signed interval, decoded as a chrono duration (Duration).
+            process_rows_to_columns!(
+                rows,
+                [u64, ::sqlx::types::BigDecimal],
+                [(::chrono::Duration, |d: ::chrono::Duration| {
+                    let nanos = d.num_nanoseconds().unwrap_or(0);
```
AndreaBozzo: Please do not turn a conversion failure into the plausible value `00:00:00`. The repository contract explicitly avoids swallowing decode/representation errors into defaults. A decoded MySQL `TIME` is bounded tightly enough that nanoseconds always fit in `i64`, so if this representation remains, use an `expect` that documents that invariant (or avoid the conversion via `MySqlTime` as suggested above). The same applies to the streaming copy.

crates/dataprof-db/tests/typed_column_decoding.rs
```diff
@@ -317,6 +318,15 @@ mod mysql {
         assert_eq!(column(columns, "amount"), ["1234.5678", ""]);
     }
 
+    #[tokio::test]
+    async fn time_of_day_decodes_even_though_it_is_not_a_date() {
+        let Some(columns) = decoded().await else {
+            return;
+        };
+
+        assert_eq!(column(columns, "t"), ["-837:59:59.123000", ""]);
```
AndreaBozzo: `MySqlTime` is converted through `to_string()` in the backend type arm. SQLx `Display` omits trailing fractional zeroes unless an explicit precision is requested. The decoded microseconds are 123000, so this renders as `-837:59:59.123`, not `.123000`; the database integration test will fail here and in the streaming assertion below. Please update both expected strings to `.123`.

AndreaBozzo (approved):
Thank you so much for this contribution, @AnandkumarMall — and especially for the patience and quick iteration through the merge conflicts and review feedback. I really appreciate it. The final solution is clean and well scoped: it uses SQLx's native MySqlTime support, avoids extra dependency and macro complexity, and adds regression coverage for batch and streaming paths, fractional values, and NULLs. Great work!

PR #658, opened 2026-09-01
fix(metrics): count identical JSON records as duplicates across a late key

### LINKED ISSUE #656: JSON: identical records are not counted as duplicates when a key appears between them
## Summary

On the JSON/JSONL path, two byte-identical records are not counted as
duplicates when a new key appears between them. The duplicate count is
silently short.

## Reproduction

`late.jsonl`:

```json
{"a": "x"}
{"a": "x", "b": "y"}
{"a": "x"}
```

Records 1 and 3 are identical, so one duplicate is expected.

```python
import dataprof as dp

report = dp.profile("late.jsonl")
print(report.quality.duplicate_rows)
```

Observed on `master` at abd4d5c, dataprof 0.11.0:

```
duplicate_rows: 0
uniqueness: {'duplicate_rows': 0, 'duplicate_rows_approximate': False,
             'high_cardinality_warning': False, 'key_column': None,
             'key_uniqueness': 100.0, 'rows_checked': 3}
```

Expected `duplicate_rows: 1`.

## Cause

`feed_json_object` (`crates/dataprof-json/src/lib.rs`) passes the
*accumulated* `known_columns` to `StreamingColumnCollection::process_record`,
which builds the row signature positionally over exactly the headers it is
given. Records read before a key is discovered are therefore signed against a
shorter header list than records read after it:

| Record | Headers at the time | Signature |
| --- | --- | --- |
| `{"a": "x"}` | `[a]` | `1:x` |
| `{"a": "x", "b": "y"}` | `[a, b]` | `1:x1:y` |
| `{"a": "x"}` | `[a, b]` | `1:x0:` |

Records 1 and 3 are the same record and hash differently, so the estimator
counts two distinct rows.

`init_column_with_missing` already backfills the *column* counters when a key
appears late, and `RowCompletenessTracker::invalidate_completed_rows` already
corrects the completeness count for the same event. The row signature is the
remaining place where a late key is not accounted for.

## Why it matters

The count is short rather than absent, so it reads as a real measurement.
`AGENTS.md` puts this in the worst bug class for a profiler: a silent error that
becomes a plausible number. The uniqueness dimension of the quality score is
computed from it.

It is data-dependent in an unhelpful way: a JSONL file whose first record
happens to carry every key is unaffected, so the bug appears and disappears with
the input rather than with the code.

## Possible directions

- Sign rows against a canonical key set rather than the discovery-order list,
  for example by sorting field names into the signature. This changes the
  signature format, so it must stay consistent with the CSV path, which
  `record_batch_analyzer.rs` deliberately matches so duplicate counts agree
  across engines.
- Re-sign is not possible after the fact: signatures are folded into the
  cardinality estimator and the records are gone. So either the signature is
  canonical from the first record, or the count is withheld for streams where a
  key arrived late, the way #650 withholds it for a merge across differing
  schemas.

The first is better if it can be done without breaking cross-engine agreement,
since withholding loses a metric that users on well-formed JSON currently get.

## Acceptance criteria

- [ ] The reproduction above reports one duplicate.
- [ ] Duplicate counts still agree between the CSV, JSON, and Arrow/Parquet
      paths on inputs all three can read.
- [ ] A test covers a key appearing after the first record, with identical
      records on both sides of its appearance.
- [ ] If withholding is chosen instead, the summary is `None`, never a short
      count.

## Likely files

- `crates/dataprof-json/src/lib.rs`
- `crates/dataprof-runtime/src/streaming_stats.rs`
- `crates/dataprof-parquet/src/record_batch_analyzer.rs` for signature parity

AndreaBozzo:
Closes #656.

## The bug

Row signatures were built positionally over whatever headers were known when
the record was read. JSON/JSONL has no header, so a key can appear at any
point and widen the schema; a record read before that point signed one field
shorter than a byte-identical record read after it.

```
{"a": "x"}              headers [a]     -> 1:x
{"a": "x", "b": "y"}    headers [a, b]  -> 1:x1:y
{"a": "x"}              headers [a, b]  -> 1:x0:
```

Rows 1 and 3 are the same record and hashed differently, so the estimator
counted two distinct rows and `duplicate_rows` reported 0 instead of 1. Short
rather than absent, so it read as a real measurement — and the uniqueness
dimension of the quality score is computed from it.

`init_column_with_missing` already backfilled the column counters for a late
key and `RowCompletenessTracker::invalidate_completed_rows` already corrected
the completeness count; the row signature was the remaining gap.

## The fix

Signatures drop trailing absent fields, which makes them independent of how
wide the schema had grown when the row was read. The trim is safe because an
absent field and an explicitly empty one are already the same null to this
profiler: it collapses only rows that hold equal values in every column they
share. It also handles a key that a later key pushes into the middle of the
schema — a row missing both signs `1:1` `0:` `0:` and trims back to the
one-column form.

On a fixed-schema path (CSV, Parquet, database, Arrow) every signature has the
same arity, so the trim applies uniformly and is injective: no count changes
there.

The length-prefixed encoding moves into a shared `RowSignature` builder that
both the incremental collection and the Arrow batch tracker use, so the two
engines cannot drift apart on how a row is signed.

## Verification

Regression tests confirmed failing before the fix and passing after:

- `dataprof-json`: `test_identical_records_count_as_duplicates_across_a_late_key`
  reproduces the issue byte for byte (0 duplicates before, 1 after), plus a
  counterpart guard that a differing row is still not merged.
- `dataprof-runtime`: two `RowSignature` cases — the late key at the end of the
  schema and a second late key that pushes the first into the middle.
- `python/tests/test_engine_parity.py`: the late-key file must report the same
  uniqueness block as the same data with the key present in every record
  (`0 == 1` before the fix).

The `RowSignature` extraction in `dataprof-parquet` is behaviour-preserving and
is *not* discriminated by the new tests; it is covered by the existing
`test_duplicate_rows_with_nulls_match_across_engines` (Rust) and
`test_duplicate_rows_with_nulls_agree_across_engines` /
`test_duplicate_rows_tracked_for_dataframe_and_arrow_inputs` (Python), which
still pass unchanged.

Cross-path spot check on the same four rows with a trailing null column —
CSV, Parquet and JSONL all report 1 duplicate over 4 rows checked.

The issue's Python repro now prints `duplicate_rows: 1`.

Commands run:

```
cargo fmt --all
cargo test --workspace
uv run pytest python/tests -q          # 1056 passed, 9 skipped
uv run ruff format / ruff check python/ .github/scripts/ .claude/skills/dataprof/scripts/
uv run ty check python/
```

### DIFF
--- a/crates/dataprof-json/src/lib.rs
+++ b/crates/dataprof-json/src/lib.rs
@@ -1663,6 +1663,44 @@ mod tests {
     }
 
     #[test]
+    fn test_identical_records_count_as_duplicates_across_a_late_key() {
+        // Records 1 and 3 are byte-identical; `b` is only discovered between
+        // them. Signing a row against whatever columns happened to be known at
+        // the time would make the two look different and report no duplicate.
+        let data = b"{\"a\": \"x\"}\n{\"a\": \"x\", \"b\": \"y\"}\n{\"a\": \"x\"}\n";
+        let cursor = Cursor::new(data.as_ref());
+        let config = JsonParserConfig::jsonl();
+
+        let (_profiles, stats, rows, _malformed, _format) =
+            analyze_json_from_reader(cursor, &config).unwrap();
+        assert_eq!(rows, 3);
+
+        let summary = stats
+            .row_duplicate_summary()
+            .expect("three rows were observed");
+        assert_eq!(summary.rows_checked, 3);
+        assert_eq!(summary.duplicate_rows, 1);
+        assert!(!summary.approximate);
+    }
+
+    #[test]
+    fn test_a_late_key_does_not_invent_duplicates() {
+        // The counterpart guard: collapsing the trailing absent field must not
+        // merge rows that differ in the columns they do share.
+        let data = b"{\"a\": \"x\"}\n{\"a\": \"x\", \"b\": \"y\"}\n{\"a\": \"z\"}\n";
+        let cursor = Cursor::new(data.as_ref());
+        let config = JsonParserConfig::jsonl();
+
+        let (_profiles, stats, _rows, _malformed, _format) =
+            analyze_json_from_reader(cursor, &config).unwrap();
+
+        let summary = stats
+            .row_duplicate_summary()
+            .expect("three rows were observed");
+        assert_eq!(summary.duplicate_rows, 0);
+    }
+
+    #[test]
     fn test_analyze_json_file_quality_report() {
         let file = write_file(r#"[{"x":1},{"x":2}]"#);
         let config = JsonParserConfig::default();
--- a/crates/dataprof-parquet/src/record_batch_analyzer.rs
+++ b/crates/dataprof-parquet/src/record_batch_analyzer.rs
@@ -11,8 +11,8 @@ use dataprof_metrics::CardinalityEstimator;
 use dataprof_metrics::analysis::inference::is_null_like_token;
 use dataprof_metrics::{RowCompletenessSummary, RowDuplicateSummary, infer_type};
 use dataprof_runtime::{
-    ColumnProfileInput, ExactNumericAggregates, RowCompletenessTracker, RowUniquenessTracker,
-    StreamReservoirSampler, TextLengths, build_column_profile,
+    ColumnProfileInput, ExactNumericAggregates, RowCompletenessTracker, RowSignature,
+    RowUniquenessTracker, StreamReservoirSampler, TextLengths, build_column_profile,
 };
 use std::collections::HashMap;
 use std::collections::hash_map::Entry;
@@ -81,8 +81,8 @@ fn observe_batch_rows(
     use arrow::datatypes::DataType as ArrowDataType;
     use arrow::util::display::{ArrayFormatter, FormatOptions};
 
-    // Null must render as an empty field ("0:"), matching the incremental
-    // engine — the default formatter would render a visible placeholder.
+    // Null must render as an empty field, matching the incremental engine —
+    // the default formatter would render a visible placeholder.
     let options = FormatOptions::default().with_null("");
     let formatters: Vec<Option<ArrayFormatter>> = batch
         .columns()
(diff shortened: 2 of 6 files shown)


AndreaBozzo:
Addressed the Copilot finding in 872c9ca, plus one thing the self-review turned up.

**The finding was right.** `crates/dataprof-python/src/columns.rs` (the ad-hoc DataFrame/dict/Arrow column path) built the length-prefixed signature inline, and its comment claimed "the same length-prefixed signature encoding as the file engines" — which this PR had just made false. It now goes through `RowSignature` too. A sweep for every `RowUniquenessTracker::observe` caller confirms that was the last one: the three producers are `streaming_stats.rs` (CSV/JSON/incremental), `record_batch_analyzer.rs` (Parquet/Arrow) and `columns.rs` (ad-hoc), all single-sourced now.

No count changes on that path — its schema is fixed, so the trailing trim applies uniformly. Verified across all five input transports on rows `(x,null) (x,null) (x,null) (z,null)`:

| input | duplicate_rows | rows_checked |
| --- | --- | --- |
| DataFrame | 2 | 4 |
| dict | 2 | 4 |
| CSV | 2 | 4 |
| Parquet | 2 | 4 |
| JSONL | 2 | 4 |

On the suggestion to keep `RowSignature` non-public: it needs to cross crate boundaries to reach `dataprof-parquet` and `dataprof-python`, and it does not enter the released API — the `dataprof` facade re-exports only `ProfileReport` and `REPORT_SCHEMA_VERSION` from `dataprof-runtime`. `pub` in an internal workspace crate is what `RowUniquenessTracker` and `RowCompletenessTracker` already are, so this follows the existing pattern.

**Separately, from the self-review:** the trim keys on an *exactly* empty value, and that is load-bearing. `is_null_like_token` — the rule that decides `null_count` — also matches `null`, `nan` and whitespace. Trimming on that broader rule would collapse a row ending in one of those against a row that ends in no field at all, fabricating duplicates instead of losing them. That was an unguarded invariant one plausible edit away from a wrong count in the opposite direction, so it now has a comment and `test_trailing_null_like_token_is_not_trimmed`, confirmed to report 2 invented duplicates when the rule is swapped and 0 with it correct.

Full suite re-run after both changes: `cargo test --workspace` clean, `cargo clippy --all --all-targets -- -D warnings` clean, 1056 Python tests pass, ruff and ty clean.

Issue #365, opened 2026-07-09
Database: decode NUMERIC/DECIMAL, temporal, UUID and BLOB columns instead of nulling them


AndreaBozzo:
## Context

0.9.0 fixed database connectors reporting every non-text column as all-null. Values are now decoded by trying concrete types in order (`String`, `i64`, `i32`, `i16`, `f64`, `f32`, `bool`) and taking the first that decodes — see `db_column_to_string!` in `crates/dataprof-db/src/connectors/common.rs`.

A value whose SQL type matches none of those arms still falls through to `None` and is recorded as a null. Documented as a known limitation in `docs/guides/database-connectors.md`.

## Types still profiling as null

| Type | Backends | Needs |
|---|---|---|
| `NUMERIC` / `DECIMAL` | Postgres, MySQL | sqlx `bigdecimal` or `rust_decimal` feature |
| `DATE`, `TIME`, `TIMESTAMP`, `TIMESTAMPTZ` | all three | sqlx `chrono` or `time` feature |
| `UUID` | Postgres | sqlx `uuid` feature |
| `BLOB` / `BYTEA` | all three | a policy decision, not a feature flag |
| unsigned `BIGINT` | MySQL | `u64` has no `Type<Postgres>` impl, so it cannot join the generic chain |

Temporal types matter most: the **Timeliness** ISO dimension (future dates, stale data ratio, temporal violations) is inert on database sources today, because every date column arrives as a null.

## Proposal

1. Enable the `chrono` sqlx feature and add temporal arms. Render as RFC 3339 so the existing string-based type inference recognises them.
2. Enable a decimal feature and add a `NUMERIC`/`DECIMAL` arm. Preserve the exact decimal string — do not route through `f64`, which defeats the point of the type.
3. Decide `BLOB` policy explicitly. Profiling bytes as text is meaningless; a `binary` column type reporting only null-count and byte-length distribution is probably right. Rendering it as a lossy string is definitely wrong.
4. MySQL unsigned integers need a backend-specific arm, since the generic chain cannot express `u64`.

Each arm must be ordered deliberately, for the reason documented on `db_column_to_string!`: `Option<String>` stays first because sqlx skips the type-compatibility check for NULL values, making it the NULL detector as well as the text arm; integers must precede `bool` because SQLite decodes `42` as `true`.

## Acceptance

- Extend `crates/dataprof-db/tests/column_decoding.rs` with a case per type.
[...]

PR #660, opened 2026-09-02
fix(metrics): accept RFC 3339 offsets, on one date grammar for every engine

### LINKED ISSUE #643: RFC 3339 timestamps profile as text: the date grammar rejects a trailing offset or Z
## Problem

A CSV or JSON column of RFC 3339 timestamps profiles as **text**, not as a date, so the Timeliness dimension assesses nothing for it and the column reports string statistics instead of temporal ones.

```
created_at
2024-01-15T10:30:00Z
2024-01-15T11:45:00Z
2024-01-15T12:00:00+02:00
```

`data_type` is `string`, `avg_length` is a number about the text, and Timeliness is absent from `assessed_dimensions()`.

## Root cause

The two grammars that decide "does this look like a date" both stop at a whole second:

- `crates/dataprof-metrics/src/analysis/inference.rs:26` — `^\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}(?:\.\d+)?$`
- `crates/dataprof-metrics/src/analysis/metrics/utils.rs:13` — `DATE_VALIDATION_REGEXES`, date-only forms

Neither accepts a trailing `Z` or a `±HH:MM` offset, which is precisely what makes a timestamp RFC 3339. `is_date_token` is the union of the two, so an offset-bearing timestamp is not a date token on any path.

## Why this matters more than it looks

RFC 3339 is the default rendering for timestamps in JSON APIs, in most export tooling, and in Parquet-to-CSV dumps. This is not an exotic format; it is the most likely form for a timestamp column arriving as text.

It also constrains the database connectors. #365 wanted to render decoded temporal columns as RFC 3339 so the existing inference would recognise them; that would have produced text columns and an inert Timeliness dimension. The connectors now render naive ISO and drop the offset, converting `TIMESTAMPTZ` to UTC first, purely because the grammar cannot accept the offset. That workaround disappears if this is fixed.

## Proposal

Extend both grammars to accept an optional RFC 3339 offset (`Z`, `+HH:MM`, `-HH:MM`, and the `HHMM` form if the validation set is meant to be permissive), and decide explicitly what an offset means downstream:

- Timeliness compares against "now"; an offset-bearing value has a well-defined instant, so the comparison gets *more* correct, not less.
- Two values with the same wall clock and different offsets are different instants. If any temporal metric compares values to each other, it must normalise before comparing rather than lexically.

## Acceptance

- [ ] A CSV column of RFC 3339 timestamps, mixed `Z` and numeric offsets, infers as a date column.
- [ ] Timeliness assesses it, with `date_values_checked` matching the non-null count.
- [ ] The two grammars agree on the new forms; `is_date_token_accepts_every_form_either_regex_set_recognizes` still holds.
- [ ] Whether the offset is preserved or normalised is decided and documented, not left to whichever code path reads it first.
- [ ] Once this lands, revisit the naive-ISO rendering in `db_column_to_string!` (#365) and decide whether the connectors should stop dropping the offset.

Found while implementing #365.

AndreaBozzo:
Closes #643.

## What was wrong

A column of RFC 3339 timestamps profiled as text — but only on some engines. Type inference ran on **two different lists of date patterns**, and they disagreed:

| | grammar | `2024-01-15T10:30:00Z` |
| --- | --- | --- |
| `infer_type` (buffered paths, Parquet fallback, Python columns) | `DATE_REGEXES` — `…T\d{2}:\d{2}:\d{2}(?:\.\d+)?$` | rejected |
| `infer_data_type_streaming` | private `DATE_PATTERN_REGEXES` — `^\d{4}-\d{2}-\d{2}T\d{2}:` | accepted by accident (unanchored) |

The same six-row CSV, measured before this change:

```
      Auto: type=Date   invalid=Some(0) timeliness_checked=6  stats=DateTime
Incremental: type=Date   invalid=Some(0) timeliness_checked=6  stats=DateTime
   Columnar: type=String invalid=None    timeliness_checked=0  stats=Text
```

That is the output contract in AGENTS.md failing: profile numbers must be identical regardless of which engine produced them.

## What changed

- **One grammar.** `looks_like_date` now delegates to `is_inferred_date_token`; `DATE_PATTERN_REGEXES` is deleted. Both entry points decide "is this a date" on the same set.
- **RFC 3339 offsets accepted.** The ISO datetime pattern takes an optional `Z`/`z` or `±HH:MM`. Those are exactly the alternatives `chrono::DateTime::parse_from_rfc3339` reads, so a value the grammar calls a date is a value the parser can turn into an instant. The ISO 8601 basic form `+0200` is deliberately excluded — chrono rejects it, and accepting it would type the column `Date` and then count every value as `invalid_date_values`.
- **Offsets normalize to UTC**, decided and documented in the `stats::datetime` module docs. `2024-01-15T23:00:00-05:00` counts as 2024-01-16 hour 04. Keeping each value's own wall clock would make two values naming the same instant compare as different, and would compare an offset-bearing value against a UTC "now" in Timeliness.

## Behavior changes

**Consistency scores rise for columns of offset-bearing timestamps.** Those values used to fail `is_date_token` and count against the column they had just been used to type. Measured on `python/tests/fixtures/dogfood/incidents.csv` (30 RFC 3339 values out of 350 checked):

| | before | after |
| --- | --- | --- |
| `data_type_consistency` | 91.43% | 100.0% |
| `quality_score` | 93.35 | 95.25 |

30/350 = 8.57%, the exact gap — every timestamp was penalized. `test_support_incident_csv_session` asserted `quality_score < 95`, which was encoding the bug; it now asserts the dimension that actually moved.

**Truncated timestamp shapes stop typing as dates.** The old unanchored pattern accepted `2024-01-15T10:` and `2024-01-15T10:30`, typed those columns `Date`, and then failed to parse a single value — empty datetime stats and a full `invalid_count`. They profile as text now, which is what they are.

**The `db_column_to_string!` naive-ISO rendering stays** (#643's follow-up item, and #365's workaround). With offsets normalized to UTC, `…T10:30:00` and `…T10:30:00+00:00` produce identical profiles, so rendering the offset would buy fidelity in the raw sample values and nothing in the numbers. The stale rationale comments saying the grammar *forces* this are corrected in `common.rs`, the connector test, and `docs/guides/database-connectors.md`.

## Acceptance criteria from #643

- [x] A CSV column of RFC 3339 timestamps, mixed `Z` and numeric offsets, infers as a date column — on all three engines
- [x] Timeliness assesses it, with `date_values_checked` matching the non-null count
- [x] The grammars agree; `is_date_token_accepts_every_form_either_regex_set_recognizes` still holds, with its example table updated
- [x] Offset handling decided (normalize to UTC) and documented
- [x] `db_column_to_string!` revisited; decision recorded above and in the code

## Verification

`tests/rfc3339_timestamps.rs` covers all three engines. Each part of the fix was reverted **separately** to confirm the tests discriminate it:

| reverted | failing tests |
| --- | --- |
| `inference.rs` (grammar) | 3 of 4 |
| `patterns.rs` (unification) | `a_timestamp_shaped_value_no_parser_accepts_stays_text` |
| `datetime.rs` (UTC) | `offsets_are_normalized_to_utc_before_the_statistics` |

Commands run:

```
cargo test --workspace                                  # 43 targets, 0 failures
cargo fmt --all --check
cargo +1.98 clippy --all --all-targets --features sqlite -- -D warnings
uv run maturin develop --release
uv run pytest python/tests/ -q                          # 1056 passed, 9 skipped
uv run ty check python/
cargo run --example {before_after_cleaning,etl_quality_gate,messy_csv_inspection}
```

The Python surface was checked against the rebuilt extension: all three engines report `date`, `invalid_count=0`, and Timeliness checking all 6 values.

### DIFF
--- a/crates/dataprof-db/src/connectors/common.rs
+++ b/crates/dataprof-db/src/connectors/common.rs
@@ -48,12 +48,16 @@ pub fn feature_not_enabled_error(db_name: &str, feature: &str) -> DataProfilerEr
 /// `Type<Postgres>`) and `BigDecimal` (PostgreSQL and MySQL, no `Type<Sqlite>`
 /// because SQLite has no decimal type).
 ///
-/// Temporal values render in the naive ISO form dataprof's date grammar
-/// accepts (`YYYY-MM-DD` and `YYYY-MM-DDTHH:MM:SS`). That grammar rejects a
-/// trailing offset or `Z`, so rendering a timestamp as true RFC 3339 would
-/// profile the column as *text* and leave the Timeliness dimension exactly as
-/// inert as the nulls did. `TIMESTAMPTZ` is converted to UTC and its offset
+/// Temporal values render in the naive ISO form `YYYY-MM-DD` /
+/// `YYYY-MM-DDTHH:MM:SS`. `TIMESTAMPTZ` is converted to UTC and its offset
 /// dropped; the instant survives, the "this is UTC" marker does not.
+///
+/// The grammar now accepts a trailing `Z` or `±HH:MM` (#643), so this is no
+/// longer forced. It stays because it is numerically indistinguishable from the
+/// alternative: dataprof normalizes an offset-bearing value to UTC before any
+/// statistic reads it, so `...T10:30:00` and `...T10:30:00+00:00` produce the
+/// same profile. Rendering the offset would buy fidelity in the raw sample
+/// values and nothing in the numbers.
 ///
 /// Still unsupported and still recorded as null: `BLOB`/`BYTEA`, which needs a
 /// binary column type rather than a decode arm.
--- a/crates/dataprof-db/tests/typed_column_decoding.rs
+++ b/crates/dataprof-db/tests/typed_column_decoding.rs
@@ -102,10 +102,10 @@ mod postgres {
             return;
         };
 
-        // Naive ISO, not RFC 3339: dataprof's date grammar accepts
-        // `YYYY-MM-DDTHH:MM:SS` and rejects a trailing offset or `Z`, so a
-        // faithful RFC 3339 rendering would profile as text and leave
-        // Timeliness exactly as inert as the nulls do.
+        // Naive ISO, not RFC 3339. Since #643 the grammar would accept the
+        // offset too, and an offset-bearing value is normalized to UTC before
+        // any statistic reads it — so the two renderings profile identically
+        // and the connectors keep converting to UTC here instead.
         assert_eq!(column(columns, "ts"), ["2024-01-15T10:30:00", ""]);
         assert_eq!(column(columns, "tstz"), ["2024-01-15T10:30:00", ""]);
         assert_eq!(column(columns, "d"), ["2024-01-15", ""]);
--- a/crates/dataprof-metrics/src/analysis/inference.rs
+++ b/crates/dataprof-metrics/src/analysis/inference.rs
@@ -27,11 +27,16 @@ static DATE_REGEXES: LazyLock<Vec<Regex>> = LazyLock::new(|| {
         // optional `Z`/`±HH:MM` offset. The offset alternatives are exactly the
         // ones `chrono::DateTime::parse_from_rfc3339` accepts, so a value this
         // pattern calls a date is a value the parser can turn into an instant.
-        // The ISO 8601 basic form (`+0200`) is deliberately absent: chrono
-        // rejects it, and accepting it here would type the column `Date` and
-        // then count every value as `invalid_date_values`.
-        Regex::new(r"^\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}(?:\.\d+)?(?:[Zz]|[+-]\d{2}:\d{2})?$")
-            .expect("BUG: Invalid hardcoded regex pattern for ISO datetime"),
+        //
+        // The offset is range-bound, not `\d{2}:\d{2}`: chrono rejects `+24:00`
+        // and `+00:60` as out of range, so a looser pattern would type the
+        // column `Date` and then count every value as `invalid_date_values` —
+        // the exact mismatch this pattern exists to avoid. The ISO 8601 basic
+        // form (`+0200`) is absent for the same reason.
+        Regex::new(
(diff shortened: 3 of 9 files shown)


AndreaBozzo:
Addressed the review. Two of three findings were real; the third I'm declining.

**Offset range — fixed.** Correct and it broke the invariant I'd written into the comment myself. `[+-]\d{2}:\d{2}` accepts `+24:00`, `+99:99`, `-00:60`; chrono rejects all three as `OutOfRange`, so such a column typed `Date` and then counted every value as `invalid_date_values`. Now `[+-](?:[01]\d|2[0-3]):[0-5]\d`, matching what `parse_from_rfc3339` actually takes (it accepts up to `+23:59`). The out-of-range forms join `+0200` in both rejection tests.

**Trim — fixed, and the divergence was real.** Measured on this branch, three identical engines, values padded with a leading space:

| input | Auto | Incremental | Columnar |
| --- | --- | --- | --- |
| `" 2024-01-15T10:30:00Z"` | String | String | **Date** |
| `"  2024-01-15"` | String | String | **Date** |

So it predates the offset work and applies to plain ISO dates too. The cause is narrower than "input normalization" generally: inside `infer_data_type_streaming`, the integer, float and boolean predicates all call `s.trim()` and the date one did not. Fixed at that call site rather than inside `looks_like_date`, so the predicate stays a pure test on the string it is given.

**Release notes — declining.** `docs/release-notes.md` is a historical record of what a shipped release contained and is regenerated at release time, so a feature branch editing it either gets overwritten or falsifies the record for a version already out. The behavior change and re-baselining guidance are in the PR description and the commit body instead. (`docs/guides/database-connectors.md` is still updated in this PR, because this change makes a claim in it factually false — that is a correction, not a release note.)

Both new fixes are pinned by tests verified to fail when that fix alone is reverted:

| reverted | failing test |
| --- | --- |
| offset range bound | `a_timestamp_shaped_value_no_parser_accepts_stays_text` |
| `.trim()` at the streaming call site | `padding_does_not_change_which_engine_calls_a_column_temporal` |

Re-ran: `cargo test --workspace`, `cargo +1.98 clippy --all --all-targets --features sqlite -- -D warnings`, `uv run pytest python/tests/ -q` (1056 passed), ruff, ty. All green.

Issue #645, opened 2026-08-31
Binary columns need a type, not a decode arm: BLOB/BYTEA profiles as null


AndreaBozzo:
## Problem

`BLOB`/`BYTEA` columns profile as entirely null on every backend. #642 decoded the other types that fell past the end of `db_column_to_string!` and deliberately left this one, because it is not a decode arm: it is a missing column type.

`docs/guides/database-connectors.md` now names it as one of two remaining gaps, so there should be a ticket behind that claim.

## Why an arm is the wrong fix

The profiler consumes columns as `Vec<String>` and re-infers types from the text. Rendering bytes as a string produces one of two bad outcomes:

- **Lossy text.** Decoding arbitrary bytes as UTF-8 mangles or rejects them, and any statistics computed from the result describe the mangling.
- **Hex or base64.** Faithful, but then length statistics measure the encoding rather than the value, cardinality counts encoded strings, and pattern detection runs over hex digits. The numbers would be real and meaningless, which is worse than the current null.

The same reasoning already applies elsewhere: #637 is the file-side version of this, where nested Parquet columns are profiled as their rendered form.

## Proposal

A binary column type that reports only what is meaningful for opaque bytes:

- null count and null ratio, as today;
- byte-length distribution (min, max, mean), named as bytes rather than "length", per the unit contract settled in #641;
- distinct count, which is well defined over bytes;
- no pattern detection, no text statistics, no numeric statistics.

Everything the profiler cannot honestly say about opaque bytes stays `None`, which is the contract the rest of the report follows.

## Scope

- A `DataType::Binary` (or equivalent) in `dataprof-core`, and the report shape for it.
- Decode arms for `Vec<u8>` per backend.
- Whether file paths ever produce it: Parquet has a `Binary` Arrow type that currently profiles through the generic text path, so this decision covers more than the database connectors.
- Serialized-report and schema implications, since a new column type is a schema-visible addition.

## Acceptance

- [ ] A `BYTEA`/`BLOB` column reports null count, byte-length statistics and distinct count, and reports nothing it cannot honestly compute.
- [ ] No text or numeric statistics are fabricated for it.
- [ ] Parquet `Binary` columns and database blobs agree, per the output contract.
- [ ] The remaining-gaps note in the connectors guide loses its `BLOB`/`BYTEA` entry.

Follow-up to #365 / #642.

Issue #637, opened 2026-08-31
Nested Parquet columns are profiled as their Arrow display string, and the JSON path renders the same value differently


AndreaBozzo:
## What happens

A Struct, List or Map column read from Parquet never reaches a typed accumulator. It falls through the type dispatch to `process_generic_array`, which renders each value with an Arrow `ArrayFormatter` and feeds the resulting display string to the text statistics, the cardinality estimator and pattern detection. The column is then reported with type `String`.

- `crates/dataprof-parquet/src/record_batch_analyzer.rs:1018` - `ArrayFormatter` output goes into `update_text_stats`, `offer_sample` and `cardinality.insert_owned`.
- `crates/dataprof-parquet/src/record_batch_analyzer.rs:1162` - the `_ => DataType::String` catch-all in the Arrow-to-dataprof type map.

So `min_length`, `max_length`, `distinct_count` and any detected patterns on a nested column describe Arrow's formatter output rather than the data.

## The cross-path part

The JSON path already has a convention for nested values, and it is a different one:

- JSON: `crates/dataprof-json/src/lib.rs:843` - `json_value_to_string` serialises objects and arrays with `serde_json::to_string`, producing `{"a":1,"b":"x"}`.
- Parquet: `arrow-cast-59.1.0/src/display.rs:1302` - a struct writes as `{a: 1, b: x}`: unquoted keys, unquoted strings, `", "` separators, and a null child rendered as the empty string.

For the record `{"a": 1, "b": "x"}` that is 15 characters on the JSON path and 12 on the Parquet path. A null child is `null` on one path and absent on the other. Same logical data, different reported numbers, decided by which file format it arrived in.

That contradicts the output contract in AGENTS.md: "profile numbers must be identical regardless of which engine or input path produced them."

## Coverage

There is no test for a nested column on the Parquet path. `StructArray` appears in the workspace only in `crates/dataprof-python/src/arrow_export.rs` (FFI export). `python/tests/test_python_api.py:316` pins the JSON-path convention for nested containers, with no Parquet counterpart.

## First step

Write the cross-path parity test before changing anything: profile the same nested record as JSON and as Parquet, assert the column profiles agree, and confirm it fails on current `master`. A guard that has never failed guards nothing.

## Options for the end state

1. Canonicalise both paths to the same compact JSON serialisation, keeping nested columns as text columns.
[...]

PR #661, opened 2026-09-02
fix(parquet): profile a column by its values, not its physical encoding

### LINKED ISSUE #647: Dictionary-encoded and view-typed Parquet columns profile as text: the fallback skips type re-inference
## Problem

The same values profile differently depending on the physical Arrow encoding the
writer chose. An ordinary `pandas.DataFrame.to_parquet` is enough to show it:

```python
import pandas as pd, dataprof

df = pd.DataFrame({"plain": [str(i % 9 + 1) for i in range(60)]})
df["cat"] = df["plain"].astype("category")
df.to_parquet("cat.parquet")

rep = dataprof.profile_file("cat.parquet")
# plain: data_type=integer  invalid_count=0     stats={min, max, mean, median, quartiles, outliers, ...}
# cat:   data_type=string   invalid_count=None  stats={min_length, max_length, avg_length}
```

Both columns hold the same sixty strings. `plain` arrives as `LargeUtf8` and is
re-inferred as `integer` with the numeric block. `cat` arrives as
`Dictionary(Int8, Utf8)` and is reported as text, with `invalid_count` absent
rather than `0`.

That is the output contract in AGENTS.md failing on a file dataprof reads today:
profile numbers must be identical regardless of which engine or input path
produced them.

## Which encodings are affected

Measured against 0.11.0 on Windows, one column per file, forty identical values
each, written by pyarrow 25.0.1 with its default `store_schema=True`:

| written as | data_type | stats block | invalid_count |
| --- | --- | --- | --- |
| `string`, `int32`, `int16`, `uint64`, `decimal128` | integer / float | numeric | 0 |
| `dictionary<int32, string>` | string | length | None |
| `dictionary<int32, int32>` | string | length | None |
| `string_view` | string | length | None |
| `binary_view` | string | length | None |
| `float16` | string | length | None |
| `time64(us)` | string | length | None |

Temporal detection goes the same way: fifty ISO dates written as `string`
profile as `date`, and the same values written as `string_view` profile as
`string`.

`binary_view` also disagrees with `binary`: the same three-byte values report
`max_length` 6 through the fallback and 10 through the `Binary` arm, because the
two paths measure different renderings of the same bytes. What a binary column
should report is #645; this issue is only about the two paths disagreeing.

The trigger is the `ARROW:schema` key-value metadata the writer embeds and the
Arrow reader honours. Writing the same `string_view` array with
`store_schema=False` round-trips as `Utf8` and profiles correctly, which places
this on the reader side rather than in Parquet. Polars 1.43 writes
`large_string`, so its files are unaffected today. pandas and pyarrow are not.

## Cause

`RecordBatchAnalyzer` dispatches on the Arrow type in three places, and the
three lists are not the same set:

- `process_array` (`crates/dataprof-parquet/src/record_batch_analyzer.rs:454-604`)
  has arms through `Decimal256` and `Duration`, then sends everything else to
  `process_generic_array`.
- `infer_data_type` (`record_batch_analyzer.rs:1140-1165`) re-runs `infer_type`
  over the collected samples only for `Utf8`/`LargeUtf8`, and returns
  `DataType::String` for everything else. The fallback does collect the right
  sample strings; they are simply never re-inferred.
- `should_track_date_matches` (`record_batch_analyzer.rs:424-434`) lists five
  types, so `exact_date_matches` stays 0 for anything that fell through.

`process_generic_array` (`record_batch_analyzer.rs:1018-1040`) has a second
problem. When `ArrayFormatter::try_new` fails it fabricates
`<{data_type}:value_{index}>` per row, which makes every row distinct: the
column then reports full cardinality and a uniqueness ratio of 1.0 for data it
could not read at all. That is a decode failure turning into a plausible number,
which AGENTS.md rules out.

The columnar CSV engine carries its own narrower copy of the same three matches
in `crates/dataprof-parquet/src/arrow_profiler.rs:443`, `:454` and `:930`. It
only ever builds `Utf8` fields, so nothing is wrong there now, but the
duplication is how the sets drifted apart in the first place.

## Proposal

- Decode the encodings that carry the same logical values as an arm that already
  exists: `Dictionary(_, _)`, `Utf8View`, `BinaryView`, `Float16`, `Time32`,
  `Time64`. A dictionary column should profile as its value type, not as its
  rendering.
- Derive the three matches from one list so they cannot drift again.
- Types with no honest text rendering (list, struct, map, binary) belong to #637
  and #645. Whatever is decided there, the fabricated `<type:value_N>` path
  should not survive it.

## Acceptance

- [ ] A dictionary-encoded column profiles identically to the same values
      written undictionaried: data_type, invalid_count, and the stats block.
- [ ] `Utf8View` agrees with `Utf8`, and `BinaryView` with `Binary`.
- [ ] Temporal and numeric re-inference run for every encoding that can carry
      them as text.
- [ ] A decode failure in the generic path is an error, not a synthetic value
      that inflates cardinality.
- [ ] A regression test writes one file per encoding and asserts profile
      equality across them, so the next Arrow type without an arm fails the
      suite instead of quietly profiling as text.

Related: #637 (nested columns rendered as their display string), #645 (binary
needs a type, not a decode arm), #547 (which surface the identical-numbers
contract governs).

AndreaBozzo:
Closes #647.

## What was wrong

The same values profiled differently depending on the physical encoding the writer chose. Measured with pyarrow 25.0.1, forty identical values per file, before this change:

```
       string: type=integer  invalid=0     min=1.0  max=9.0   ← numeric block
 large_string: type=integer  invalid=0     min=1.0  max=9.0
     dict_str: type=string   invalid=None  minlen=1 maxlen=1  ← length block
  string_view: type=string   invalid=None  minlen=1 maxlen=1
     dict_int: type=string   invalid=None  minlen=1 maxlen=1
      float16: type=string   invalid=None  minlen=1 maxlen=1
       binary: type=string   invalid=None  minlen=10 maxlen=10
  binary_view: type=string   invalid=None  minlen=6  maxlen=6  ← disagrees with binary
```

`pandas.DataFrame.to_parquet` on a categorical column is enough to hit it. `RecordBatchAnalyzer` dispatched on the Arrow type in **three** places — `process_array`, `infer_data_type`, `should_track_date_matches` — and the three lists were not the same set.

After:

```
       string: integer  invalid=0  |  dict_str:    integer  invalid=0
 large_string: integer  invalid=0  |  string_view: integer  invalid=0
        int32: integer  invalid=0  |  dict_int:    integer  invalid=0
      float16: float    invalid=0  |  binary/binary_view: both maxlen=10
```

## What changed

- **`logical_arrow_type` strips the encoding once.** `Dictionary(_, v)` → `v`, `RunEndEncoded(_, v)` → `v`, `Utf8View` → `Utf8`, `BinaryView` → `Binary`, `Float16` → `Float32`. The column is normalized at construction and the array cast on entry, so all three matches see one type.
- **The two remaining matches derive their fallback instead of enumerating it.** `infer_data_type` re-infers over the collected samples rather than declaring `String`; date tracking is on by default. The next Arrow type without an arm is right by construction instead of silently text with a date count that was never taken.
- **`process_generic_array` no longer fabricates `<type:value_N>` per row** when a formatter cannot be built. That made every row distinct, so a column dataprof could not read at all reported full cardinality and a uniqueness ratio of 1.0. It is an error now.
- **The columnar CSV engine's copy** of the same three matches gets the same treatment, plus its `array_value_to_string(...).unwrap_or_else(|_| "<type:…>")`, which swallowed a decode error into a plausible value. It only builds `Utf8` fields today, but the duplication is how the sets drifted apart.

## A hazard found while verifying, and guarded

The obvious version of the re-inference fix — `_ => infer_type(samples)` for everything — is **wrong**, and I only caught it by testing a type the issue does not mention. `FixedSizeBinary` formats as hex, so re-inferring reads the three bytes `abc` back as the integer **616263** and reports a numeric column with that mean:

```
FixedSizeBinary -> type=Integer stats=Numeric(min: 616263.0, max: 616263.0, mean: 616263.0, …)
```

That is a decode artifact turned into a plausible number — the failure class AGENTS.md singles out as the worst for a profiler. Both fallbacks are therefore gated on `samples_are_text_values`, which excludes the binary families for exactly this reason, and `a_binary_rendering_is_never_re_inferred_as_the_number_it_spells` pins it.

## Acceptance criteria from #647

- [x] A dictionary-encoded column profiles identically to the same values written undictionaried — data_type, invalid_count, and the stats block
- [x] `Utf8View` agrees with `Utf8`, `BinaryView` with `Binary`
- [x] Temporal and numeric re-inference run for every encoding that can carry them as text — and *only* those, per the hex finding above
- [x] A decode failure in the generic path is an error, not a synthetic value that inflates cardinality
- [x] A regression test writes one file per encoding and asserts profile equality across them

Out of scope, unchanged: what a binary column should report (#645) and nested types (#637). This PR only makes the two paths agree on whatever those decide.

## Verification

`tests/parquet_encoding_parity.rs` compares a full profile string — type, counts, invalid, unique, and the whole stats block — so a divergence anywhere fails, not only in fields a test remembered to name.

Reverting the parts separately:

| reverted | result |
| --- | --- |
| `logical_arrow_type` normalization | `half_precision…` and `the_binary_layouts_agree…` fail |
| re-inference fallback | **no test fails** — see below |

Being explicit, per the repo's own rule about not implying coverage: **the suite does not discriminate the re-inference fallback on its own.** With normalization in place, every encoding in the test is cast to a type that has a native arm, so nothing reaches the fallback. It is kept because it is what makes a *future* Arrow type without an arm correct rather than accidentally correct, and the `FixedSizeBinary` test constrains its shape. The `a_time_of_day_is_text…` case exercises the fallback path but would pass either way.

Commands run:

```
cargo test --workspace --features parquet          # 0 failures
cargo fmt --all
cargo +1.98 clippy --all --all-targets --features parquet,sqlite -- -D warnings
uv run maturin develop --release
uv run pytest python/tests/ -q                     # 1056 passed, 9 skipped
cargo run --example {before_after_cleaning,etl_quality_gate,messy_csv_inspection}
```

Also re-checked end to end against real pyarrow-written files for all ten encodings above.

### DIFF
--- a/crates/dataprof-parquet/src/arrow_profiler.rs
+++ b/crates/dataprof-parquet/src/arrow_profiler.rs
@@ -454,38 +454,25 @@ impl ColumnAnalyzer {
         self.sample_values.offer(value);
     }
 
-    /// Whether this column's sample values are its own values written as text.
+    /// Whether this column's text rendering is an *encoding* of its values
+    /// rather than the values themselves.
     ///
-    /// Kept identical to the parquet analyzer's predicate, which carries the
-    /// full rationale — including why the binary families are excluded.
-    fn samples_are_text_values(&self) -> bool {
+    /// The same predicate as the parquet analyzer's, which carries the full
+    /// rationale. It is deliberately not a list of which types reach the text
+    /// fallback: the match arms above already answer that, and this analyzer's
+    /// arms are not the parquet one's — `Int16`, `UInt*` and `Decimal*` fall
+    /// through to text here and do not there.
+    fn renders_values_as_encoded_bytes(&self) -> bool {
         use arrow::datatypes::DataType as ArrowDataType;
 
-        !matches!(
+        matches!(
             &self.data_type,
-            ArrowDataType::Null
-                | ArrowDataType::Boolean
-                | ArrowDataType::Float64
-                | ArrowDataType::Float32
-                | ArrowDataType::Int64
-                | ArrowDataType::Int32
-                | ArrowDataType::Int16
-                | ArrowDataType::Int8
-                | ArrowDataType::UInt64
-                | ArrowDataType::UInt32
-                | ArrowDataType::UInt16
-                | ArrowDataType::UInt8
-                | ArrowDataType::Decimal128(_, _)
-                | ArrowDataType::Decimal256(_, _)
-                | ArrowDataType::Duration(_)
-                | ArrowDataType::Binary
-                | ArrowDataType::LargeBinary
-                | ArrowDataType::FixedSizeBinary(_)
+            ArrowDataType::Binary | ArrowDataType::LargeBinary | ArrowDataType::FixedSizeBinary(_)
         )
     }
 
     fn should_track_date_matches(&self) -> bool {
-        self.samples_are_text_values()
+        !self.renders_values_as_encoded_bytes()
     }
 
     fn process_array(&mut self, array: &dyn Array) -> Result<(), DataProfilerError> {
@@ -1011,7 +998,9 @@ impl ColumnAnalyzer {
             // same shared inference every other engine types its text columns
             // with. Where the rendering is an encoding rather than the value,
             // the column stays text.
-            _ if self.samples_are_text_values() => infer_type(self.sample_values.samples()),
+            _ if !self.renders_values_as_encoded_bytes() => {
+                infer_type(self.sample_values.samples())
+            }
             _ => DataType::String,
         }
     }
--- a/crates/dataprof-parquet/src/arrow_profiler.rs
+++ b/crates/dataprof-parquet/src/arrow_profiler.rs
@@ -423,7 +423,7 @@ struct ColumnAnalyzer {
 impl ColumnAnalyzer {
     fn new(data_type: &arrow::datatypes::DataType) -> Self {
         Self {
-            data_type: data_type.clone(),
+            data_type: crate::record_batch_analyzer::logical_arrow_type(data_type),
             total_count: 0,
             null_count: 0,
             cardinality: CardinalityEstimator::new(),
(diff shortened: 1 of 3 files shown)


AndreaBozzo:
Addressed all three. Each was correct, and the first changed the design.

**Arrow-profiler exclusions vs. actual dispatch — fixed by deleting the list.** Confirmed: that analyzer's `process_decoded_array` has arms for `Null, Float64, Float32, Int64, Int32, Utf8, LargeUtf8, Boolean, Date32, Date64, Timestamp, Binary|LargeBinary` — so `Int16`, `Int8`, `UInt*`, `Decimal*` and `Duration` fall through to text there and do not in the parquet analyzer. Listing them in a shared predicate blocked re-inference for samples that genuinely were text: the same drift this PR removes, reintroduced one level up.

Rather than sync two lists or build dispatch metadata, the list is gone. Which types reach the fallback is already answered by the match arms above it, per analyzer, so the predicate now carries only the part that cannot be derived — whether the rendering is an *encoding* rather than the value:

```rust
fn renders_values_as_encoded_bytes(&self) -> bool {
    matches!(self.data_type, Binary | LargeBinary | FixedSizeBinary(_))
}
```

`should_track_date_matches` follows the same predicate instead of its own list. A numeric or boolean rendering is never a date token, so counting it costs a cheap byte check and yields the same zero a skip would have — verified, no number moves — while a type this build has never seen gets an honest count instead of one that was never taken.

**Run-end encoding — tested, and the test discriminates.** You're right that it was unverified. It can't be reached through a file (Parquet has no run-end encoding), but it reaches `RecordBatchAnalyzer` directly, which is public and is what the Python Arrow import path feeds — so the test lives there.

Worth reporting how it went: a run of ordinary integers **passes with the arm removed**, so that version would have been a test that proves nothing. It needed a run of `NaN`, where the native float arm counts NaN as null and the generic formatter renders it as the string `"NaN"`. Without the cast:

```
null_count:    0  →  30
unique_count:  2  →  1
invalid_count: 30 →  0
```

**Parity helper — now compares the whole profile.** Correct that `patterns`, `type_homogeneity` and `unique_count_is_approximate` were omitted while the helper claimed otherwise, and that this change moves exactly those. It is `format!("{profile:?}")` now. All six cases still pass, so the encodings agree on everything, not just the named fields.

---

Restating the coverage limit from the description, since it survives these changes: **the suite still does not discriminate the re-inference fallback on its own.** With normalization in place every encoding under test reaches a native arm, so nothing hits it. It stays because it makes a future Arrow type without an arm correct by construction, and the `FixedSizeBinary` test constrains its shape.

Re-ran: `cargo test --workspace --all-features`, `cargo +1.98 clippy --all --all-targets --features parquet,sqlite -- -D warnings`, `uv run pytest python/tests/ -q` (1056 passed), the three examples, and the ten-encoding pyarrow check end to end. All green.

PR #666, opened 2026-09-03
fix(python): profile a zero-row Arrow source over its declared columns

### LINKED ISSUE #664: Zero-row Arrow sources are refused while a header-only CSV profiles as 0 rows
## What happens

A zero-row source is analyzed on the file paths and refused on the Arrow paths.

```python
import dataprof, pandas as pd, polars as pl, pyarrow as pa

# header-only CSV
dataprof.profile("header_only.csv")            # rows=0, columns=2

# empty JSON array
dataprof.profile("empty.json")                 # rows=0, columns=0

# the same emptiness, in memory
dataprof.profile(pa.table({"a": pa.array([], type=pa.int64())}))
dataprof.profile(pd.DataFrame({"a": pd.Series([], dtype="int64")}))
dataprof.profile(pl.DataFrame({"a": []}))
# ValueError: Arrow source is empty: it holds no record batches
```

The CSV case is the sharper one: it has a declared schema and reports two
columns over zero rows, which is the right answer. The Arrow table has a
declared schema too and raises instead.

## Why it is a contract problem, not a preference

`AGENTS.md`: "profile numbers must be identical regardless of which engine or
input path produced them." Here one path produces a profile and another
produces an exception for the same logical input.

It also sits against the repo's `None`/empty rule. A zero-row table with a
schema is "analyzed, nothing found", which is the empty case, not the absent
one. Refusing it makes emptiness unrepresentable through the Arrow entry
points, so a caller profiling a filtered frame has to special-case the filter
that matched nothing, which is exactly the case they were profiling to detect.

## Where it comes from

`import_batches_from_pyarrow` (`crates/dataprof-python/src/arrow_export.rs`)
raises when `to_batches()` returns an empty list. pyarrow returns no batches
for a zero-row table, so the schema is discarded along with the rows.

The guard is old. #662 unified its wording across the three Arrow paths, which
is why they now agree with each other and disagree with the file paths as a
group. `python/tests/test_dataframe_chunk_parity.py::test_empty_frames_are_refused_the_same_way_on_every_path`
pins current behaviour and points here.

## Outcome

A zero-row Arrow source profiles as zero rows over its declared columns, the
way a header-only CSV does.

The schema is available even when `to_batches()` is empty (`Table.schema`, and
the C Data Interface schema capsule), so the columns can be reported with the
counts they deserve.

## Design constraints

- Zero-row columns report `total_count` 0 and whatever the existing zero-row
  handling on the CSV path produces for null percentage and the rest. Copy that
  behaviour rather than inventing a second convention; `null_percentage` is
  already nullable for exactly this case.
- Absent stays distinct from empty. A statistic that cannot be computed over
  zero values is `None`, never `0.0`.
- A genuinely unusable source, one with no schema at all, should still error.
  Empty and malformed are different answers.
- `max_rows=0` already produces an empty slice that keeps its schema
  (`limit_batches`), so the report shape for zero analyzed rows exists; this
  should produce the same shape.

## Acceptance criteria

- [ ] A zero-row `pyarrow.Table`, pandas DataFrame and polars DataFrame each
      profile as 0 rows over their declared columns.
- [ ] A cross-path test asserts a header-only CSV and the equivalent zero-row
      Arrow table produce the same column profiles.
- [ ] Zero-row statistics are absent rather than zero where they are undefined.
- [ ] A source with no schema still raises, with a test separating that from
      the empty case.
- [ ] `test_empty_frames_are_refused_the_same_way_on_every_path` is replaced by
      its positive counterpart rather than deleted.
- [ ] The tests fail against the code before the change.

AndreaBozzo:
Closes #664.

## What changed

A zero-row source was analyzed on the file paths and refused on the Arrow ones:

```python
dataprof.profile("header_only.csv")   # rows=0, columns=2
dataprof.profile("empty.json")        # rows=0, columns=0
dataprof.profile(pa.table({"a": pa.array([], type=pa.int64())}))
# ValueError: Arrow source is empty: it holds no record batches
```

`import_batches_from_pyarrow` raised when `to_batches()` came back empty, and
pyarrow returns no batches at all for a zero-row Table, so the schema was
discarded along with the rows. The schema is still reachable over the C Data
Interface, so it is now imported and turned into an empty `RecordBatch`.

The existing zero-row handling downstream does the rest — it was already
exercised by `max_rows=0`, which is why this is a guard removal rather than a
new code path: `total_count` 0, `null_count` 0, and `null_percentage` and
`uniqueness_ratio` absent rather than 0.0.

The schema lives in two places depending on the producer: pyarrow puts
`__arrow_c_schema__` on `Table.schema` and not on the Table, while an object
written against the PyCapsule interface alone carries it directly. Both are
tried. A producer that offers neither is still refused, with a message that now
says why — nothing names its columns, so there is no profile to produce.

The two in-memory Python paths (`profile({"a": []})`, `profile([])`) already
handled emptiness correctly; the Arrow paths were the last holdout.

## Verification

New `python/tests/test_zero_row_sources.py`. Run against the unfixed extension
first: **7 of 8 failed**. The one that passed is
`test_zero_row_record_batch_profiles_as_zero_rows` — the RecordBatch arm never
went through the batch-list guard, and it is in the file so the Table arm is
measured against a path that was already correct rather than against nothing.

The cross-path test compares the Arrow report against a header-only CSV over
the fields emptiness decides (counts and ratios), not the whole column section:
the CSV path infers types from values and a header-only file has none, so it
types both columns as text where the Arrow schema declares one `int64`. A
declared type surviving is the point of carrying the schema.

`test_empty_frames_are_refused_the_same_way_on_every_path` pinned the old
refusal. What it guarded — one answer from all three library paths, which share
an importer — is unchanged, so it now asserts on the report they produce
instead of the error they used to raise, and is renamed accordingly. Its
columns are typed explicitly on all three sides, because an untyped polars
frame exports as an Arrow `null` column: a different declared schema, not a
different answer.

Commands run:

```
cargo fmt --all
cargo clippy --all --all-targets -- -D warnings     # clean
cargo test -p dataprof-python                       # clean
uv run maturin develop
uv run pytest python/tests/ -q                      # 1075 passed, 9 skipped
uv run ruff format python/ ; uv run ruff check python/
uv run ty check python/                             # clean
```

Local toolchain is 1.97; CI pins 1.98, so clippy is worth re-reading there.

## Left out, deliberately

A zero-row **numeric** column reports `stats: {min: 0.0, max: 0.0, mean: 0.0,
std_dev: 0.0, variance: 0.0}` — fabricated numbers over no values, against the
second design constraint in #664 ("a statistic that cannot be computed over
zero values is `None`, never `0.0`").

It is pre-existing and independent of this change: `profile(table, max_rows=0)`
produces it on today's `master`. Fixing it means either making
`NumericStats::{min,max,mean,std_dev,variance}` optional — a breaking change to
the Rust and Python surfaces and to the serialized schema, affecting every
engine and every all-null column — or emitting `ColumnStats::None` for a column
with no analyzed values, which changes the Parquet and CSV paths too. Either is
a second behaviour change; filing it separately.

### DIFF
--- a/crates/dataprof-python/src/arrow_export.rs
+++ b/crates/dataprof-python/src/arrow_export.rs
@@ -980,12 +980,25 @@ fn import_batches_from_pyarrow(
         let batch_list: Vec<Py<PyAny>> = batches.extract()?;
 
         if batch_list.is_empty() {
-            // Worded for the source, not for pyarrow: pandas and polars frames
-            // reach this same guard after converting, and "Table is empty" is
-            // not what those callers passed in.
-            return Err(PyValueError::new_err(
-                "Arrow source is empty: it holds no record batches",
-            ));
+            // A zero-row source is analyzable, not unusable. It carries a
+            // declared schema, and the file paths already profile that shape:
+            // a header-only CSV reports 0 rows over its declared columns rather
+            // than raising. Refusing it here made emptiness unrepresentable
+            // through the Arrow entry points, so a caller profiling a filtered
+            // frame had to special-case the filter matching nothing, which is
+            // the case they were profiling to detect.
+            //
+            // pyarrow returns no batches at all for a zero-row Table, so the
+            // columns have to come from the schema instead of from a batch.
+            let schema = import_schema_from_pycapsule(obj)?.ok_or_else(|| {
+                // Worded for the source, not for pyarrow: pandas and polars
+                // frames reach this same guard after converting, and "Table is
+                // empty" is not what those callers passed in.
+                PyValueError::new_err(
+                    "Arrow source is empty: it holds no record batches and exposes no schema, so its columns cannot be reported",
+                )
+            })?;
+            return Ok(vec![RecordBatch::new_empty(Arc::new(schema))]);
         }
 
         // Every batch, not just the first. A chunked Table profiled as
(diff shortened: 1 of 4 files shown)


AndreaBozzo:
Filed the deliberate omission as #667 — a numeric column with no analyzed values reporting `min/max/mean` of 0.0. Reproducible on `master` today via `max_rows=0`, so it is independent of this PR rather than introduced by it.

AndreaBozzo:
Both review points were right; addressed in 3ecadb5.

**Fabricated zero-value statistics.** Fixed rather than deferred. What settled
it is that the zero-row profile contradicted *itself*: a zero-row string or date
column already reported absent statistics, because both arms fall through to
`ColumnStats::None` on an empty sample. Only numeric and boolean fabricated. So
this is not a second behaviour change, it is the same one finished.

Done in `build_column_profile`, the single canonical construction path, keyed on
`total_count == 0`. All three zero-row entry points now agree:

| input | before | after |
| --- | --- | --- |
| zero-row Arrow table | `{min: 0.0, max: 0.0, mean: 0.0, std_dev: 0.0, variance: 0.0}` | absent |
| zero-row Parquet file | same | absent |
| `max_rows=0` | same | absent |

Boolean `true_ratio: 0.0` went the same way. Counts are untouched — "0 of 0
values were invalid" is a fact about an analyzed column, not a statistic over
nothing.

The guard is keyed on `total_count`, not on the parsed count, so it covers
exactly the empty case. A column with *rows* but no parsed values is genuinely
wider: `NumericStats::{min,max,mean,std_dev,variance}` are non-optional `f64`,
so answering `None` there means either making them optional (breaking the Rust
API, the PyO3 accessors and the serialized schema) or dropping the whole stats
block for an all-null numeric column. That is #667, now narrowed to just that
case.

**Untested direct-schema branch.** Correct, and the gap was real — every test
took the `obj.schema` branch, because pyarrow puts `__arrow_c_schema__` on
`Table.schema` and not on the Table. `test_a_producer_carrying_the_schema_itself_keeps_its_columns`
covers the other branch with an empty producer carrying the capsule directly.

Writing it turned up a detail worth recording: the custom arm of
`convert_dataframe_to_batches` gates on `__arrow_c_array__` before importing, so
an empty producer has to declare that method even though it is never called —
with no batches there is no array to hand over, and the schema is what carries
the columns.

**Verification.** `test_zero_row_columns_carry_no_statistics_at_all` was run
against the branch *with* the schema import and *without* the stats change, and
failed there — so the two parts of this PR are discriminated separately rather
than one masking the other. `test_a_single_row_still_reports_its_statistics`
guards the half a suppression fix gets wrong: a genuine `min` of 0.0 and a
genuine `true_ratio` of 0.0 must survive.

Full suites after the change: 1078 Python passed / 9 skipped, `cargo test
--workspace` clean, `cargo clippy --all --all-targets -D warnings` clean, ruff
and ty clean.

Issue #694, opened 2026-09-07
AgentGuard redacts host paths from errors, then emits the absolute path in llm_context()


AndreaBozzo:
## Summary

`AgentGuard` redacts host paths from every rejection, then emits the absolute
host path of every file it successfully profiles. `llm_context()` — the method
whose whole purpose is producing text safe to hand to a model — begins with the
resolved path, including the OS username and the sandbox root's location on
disk.

Filing this to settle the boundary rather than to claim a broken promise: the
docstrings on `to_llm_context()` promise redaction of *raw cell values*, and the
module docstring scopes its host-path guarantee to *error messages*. Both are
kept. The gap is that the success path was never given the same treatment as the
failure path, on a surface whose stated job is to sit between untrusted model
input and real files.

## Reproduction

```python
from dataprof.agent import AgentGuard, SandboxPolicy

guard = AgentGuard(SandboxPolicy(roots=["/srv/data"]))

# failure: redacted to a sandbox-relative name
try:
    guard.profile("nope.csv")
except Exception as e:
    print(guard.sanitize_error(e))

# success: absolute host path, straight into model context
print(guard.llm_context(guard.profile("ok.csv")))
```

Observed on `master` at 907e919, dataprof 0.11.0 (0.12 development):

```
no such file inside the sandbox: nope.csv

dataset: C:\Users\andre\AppData\Local\Temp\claude\...\dog\sandbox\ok.csv (file)
rows: 2 | columns: 2 | quality: 100.0/100
caveat: low sample size, quality metrics are unreliable

schema (2):
- a: integer
- b: string
```

The same absolute path is on `report.source`, `to_dict()["source"]` and
`to_markdown()`, so any of the agent-facing renderings carries it.

## Why the asymmetry looks unintended

The module is explicit that it exists as an enforcement point, and
`sanitize_error()` reduces every non-guard exception to its type on the
reasoning that an arbitrary error "may carry an absolute path or a fragment of
the file that failed to parse, and neither may reach the model."

That reasoning applies to the path in `llm_context()` too — it reaches the model
on every successful call rather than only on an unlucky one. The guard already
knows the sandbox root and already renders paths relative to it in errors, so
the information needed to redact is on hand.

What leaks is the sandbox root's absolute location, the OS username, and the
host's directory layout. Not the file contents, and not anything outside the
sandbox — but the boundary's design elsewhere treats host paths as
non-disclosable.

## Existing tests

[...]

Issue #693, opened 2026-09-07
infer_schema and analyze_structure report Parquet physical encoding, not the type profile() infers


AndreaBozzo:
## Summary

`infer_schema()` and `analyze_structure()` report a Parquet column's **physical
Arrow type**, while `profile()` reports the type inferred from its **values**.
The two disagree about the same column of the same file, and `infer_schema()` on
a Parquet file disagrees with `infer_schema()` on the identical data as CSV.

This is the bug class #661 ("profile a column by its values, not its physical
encoding") fixed on the profiling path, still present on the partial-analysis
path. It is a sibling of #552 — same surface, same shape (partial analysis
disagreeing with the full profiler on one file), found the same way.

## Reproduction

A Parquet file whose columns are physically `string` but semantically dates,
integers and booleans — the shape you get from any writer that did not type its
input, which is common for exported data.

```python
import dataprof as dp, pyarrow as pa, pyarrow.parquet as pq, csv

tbl = pa.table({
    "when":     ["2026-01-02T03:04:05Z"] * 50,
    "ident":    [f"ORD-{i}" for i in range(50)],
    "numish":   [str(i * 3) for i in range(50)],
    "boolish":  ["true", "false"] * 25,
    "real_int": list(range(50)),
})
pq.write_table(tbl, "typed.parquet")
with open("typed.csv", "w", newline="", encoding="utf-8") as f:
    w = csv.writer(f); w.writerow(tbl.column_names)
    for i in range(50):
        w.writerow([tbl.column(c)[i].as_py() for c in tbl.column_names])

prof = dp.profile("typed.parquet")
schema = {c["name"]: c["data_type"] for c in dp.infer_schema("typed.parquet").columns}
for c in tbl.column_names:
    print(f"{c:9} profile={prof[c].data_type:8} infer_schema={schema[c]}")
```

Observed on `master` at 907e919, dataprof 0.11.0 (0.12 development):

```
column     profile(pq)  infer(pq)    profile(csv)  infer(csv)   agree?
when       date         string       date          date         MISMATCH
ident      string       string       string        string       ok
numish     integer      string       integer       integer      MISMATCH
boolish    boolean      string       boolean       boolean      MISMATCH
real_int   integer      integer      integer       integer      ok
```

Three of five columns disagree. All four answers agree for CSV, JSON and JSONL;
Parquet is the only format where the fast path and the full path differ.

`analyze_structure()` inherits it through the same helper:

```python
[...]

Issue #692, opened 2026-09-07
Async stop conditions mark a fully-read source as truncated: the #460 fix landed on sync only


AndreaBozzo:
## Summary

On the async path, a stop condition that fires on the chunk which also finishes
the source marks a fully-read profile as truncated. The report profiles every
row, then reports `source_exhausted=False` and a `truncation_reason` — so a
caller cannot distinguish a complete profile from a bounded one, which is the
purpose of that metadata.

This is the defect #460 fixed on the sync path, still present on the async path.
#460's acceptance criterion "a condition firing on the final chunk does not mark
an exhausted source as truncated" holds for sync and fails for async.

Two conditions are affected: `max_bytes(n)` for any `n` below the source size,
and `schema_stable(n)`. `max_rows` is correct on both paths, and the sync path is
correct for every condition tested.

## Reproduction

Any CSV works; this used a 45,101-byte, 500-row file.

```python
import asyncio, dataprof as dp
from dataprof.asyncio import profile_bytes

data = open("orders.csv", "rb").read()          # 45,101 bytes, 500 rows
SC = dp.StopCondition

async def main():
    for label, cond in [
        ("max_bytes(100)",    SC.max_bytes(100)),
        ("max_bytes(10000)",  SC.max_bytes(10_000)),
        ("schema_stable(50)", SC.schema_stable(50)),
    ]:
        s = dp.profile("orders.csv", stop_condition=cond)
        a = await profile_bytes(data, format="csv", stop_condition=cond)
        print(f"{label:18} sync  rows={s.rows} exhausted={s.source_exhausted} reason={s.truncation_reason}")
        print(f"{label:18} async rows={a.rows} exhausted={a.source_exhausted} reason={a.truncation_reason}")

asyncio.run(main())
```

Observed on `master` at 907e919, dataprof 0.11.0 (0.12 development):

```
max_bytes(100)     sync  rows=500 exhausted=True  reason=None
max_bytes(100)     async rows=500 exhausted=False reason=max_bytes(100)
max_bytes(10000)   sync  rows=500 exhausted=True  reason=None
max_bytes(10000)   async rows=500 exhausted=False reason=max_bytes(10000)
schema_stable(50)  sync  rows=500 exhausted=True  reason=None
schema_stable(50)  async rows=500 exhausted=False reason=stop_condition(schema_stable(50))
```

`rows=500` is the whole file in every row above. Nothing was truncated: the
async report's column statistics are identical, field for field, to an
[...]

PR #695, opened 2026-09-07
docs(python): document column projection

AndreaBozzo:
## What

Documents our nice new `columns` — column projection — in the Python API guide.

It shipped in #655 and reached the type stubs, `Profiler.columns()` and
`ProfilerConfig`, but never the guide: the documented `dp.profile()` signature
omitted the parameter and no section described it. A user reading
`docs/python/README.md` had no way to learn the feature exists.

## Why these particular details

Projection is easy to guess wrong in ways that quietly change results, so the
new section records the semantics rather than just naming the parameter:

- it selects columns and never rows, and each retained column reports exactly
  what it reports unprojected;
- the report keeps **source** order, not the order the names were passed in;
- unknown and repeated names raise instead of being ignored;
- `columns=[]` (project onto nothing) is distinct from `columns=None` (no
  projection);
- quality is assessed over the projection only, so `quality_score` answers a
  narrower question than the whole-dataset score — the one that would otherwise
  surprise someone comparing a projected run against a full one;
- semantic hints resolve against the projection, so a hint naming an excluded
  column raises.

`ProfilerConfig(columns=...)` is described as reaching `analyze_database_async()`
rather than the file paths, since `dp.profile()` takes no `config=`.

## Verification

Documentation only — no source change. Every factual claim in the new section
was checked against the built extension rather than read off the stubs; 15/15
verified, including the code sample's literal output, projection on all ten
input transports and both engines, source-order preservation with names passed
in reverse, both `ValueError` paths, the `[]` / `None` distinction, and stat
parity between projected and unprojected profiles across CSV, JSONL, JSON and
Parquet (0 field differences).

```
uv run pytest python/tests/ -q                    # 1274 passed, 9 skipped
uv run pytest python/tests/test_agent_docs_sync.py \
              python/tests/test_public_surface.py \
              python/tests/test_wheel_feature_single_source.py -q   # 54 passed
```

Found while dogfooding the 0.12 surface; the reproduced defects from that pass
are filed separately as #692, #693 and #694.

### DIFF
--- a/docs/python/README.md
+++ b/docs/python/README.md
@@ -102,6 +102,7 @@ dp.profile(
     progress_interval_ms=None,       # int -- ms between progress events
     metrics=None,                    # list[str] -- "schema", "statistics", "patterns", "quality"
     quality_dimensions=None,         # list[str] -- subset of dimensions to compute
+    columns=None,                    # list[str] -- profile only these columns
     locale=None,                     # str -- "CA"|"DE"|"FR"|"GB"|"IT"|"US"
     positive_columns=None,           # list[str] -- columns expected to be non-negative
     identifier_columns=None,         # list[str] -- semantic IDs, not measures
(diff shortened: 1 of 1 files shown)


AndreaBozzo:
Both points verified against the built extension and addressed in 7aa0f25.

**1. Async entry points.** Correct. `profile_file`, `profile_bytes` and
`profile_url` each build `ProfilerConfig(**kwargs)`, so all three take
`columns`; confirmed by profiling through `asyncio.profile_file` and
`asyncio.profile_bytes` with a projection and getting the projected report back.
The entry-point list now names the async module.

**2. Withheld dimensions.** Correct, and larger than described. Projection does
not narrow `completeness` and `uniqueness` — it withholds them, so both read
`None` and drop out of `assessed_dimensions()`. `ReportAssembler::with_analysis_options`
does this deliberately: those dimensions carry row-level fields
(`complete_records_ratio`, `total_cells`, `duplicate_rows`, `rows_checked`) that
would otherwise be published as projected measurements under full-row names.

Two consequences the section now also records, both found while checking this:

- the overall score averages fewer dimensions and is usually *higher* than the
  unprojected score for the same columns (96.94 vs 94.27 on the fixture),
  because `completeness` is the dimension that penalizes nulls — so projected
  and whole-dataset scores are not comparable;
- a projection combined with `quality_dimensions` naming only those two leaves
  no assessable dimension, and the report then carries no quality object at all
  (`report.quality is None`, `quality_score is None`).

All 31 factual claims in the section are re-verified against the extension after
these edits, and the doc-sync, public-surface and wheel-feature tests pass.

PR #708, opened 2026-09-10
fix(metrics): scope and enforce the cross-engine numeric equality contract

### LINKED ISSUE #547: Settle which surface the identical-numbers contract governs: raw attributes drift across paths, serialized reports do not
## Summary

`AGENTS.md` states the output contract as: profile numbers must be identical
regardless of which engine or input path produced them. On raw attribute access
that is not quite true. The same 5,000-row dataset profiled through sixteen input
paths yields three distinct values for `mean`, `std_dev`, `variance`,
`coefficient_of_variation`, and `avg_length`, differing in the last two or three
significant digits.

The serialized report is unaffected: `to_json()` rounds these fields to four
decimal places and every path agrees exactly. Only `ColumnProfile.mean` and
friends expose the difference.

This is filed to settle which surface the contract governs, not because any
number is wrong.

## Reproduction

```python
import dataprof as dp
import pandas as pd
from pathlib import Path

F = Path("fixtures")
cases = {
    "csv.auto":     (F / "customers.csv", {}),
    "csv.columnar": (F / "customers.csv", {"engine": "columnar"}),
    "csv.bytes":    ((F / "customers.csv").read_bytes(), {"format": "csv"}),
    "parquet":      (F / "customers.parquet", {}),
    "pandas":       (pd.read_parquet(F / "customers.parquet"), {}),
}
for k, (src, kw) in cases.items():
    c = dp.profile(src, **kw).column_profiles["mrr_eur"]
    print(f"{k:14} mean={c.mean!r:22} std_dev={c.std_dev!r}")
```

```
csv.auto       mean=501.8635642933049      std_dev=288.7825168765521
csv.columnar   mean=501.8635642933046      std_dev=288.78251687655325
csv.bytes      mean=501.8635642933049      std_dev=288.78251687655177
parquet        mean=501.8635642933046      std_dev=288.78251687655325
pandas         mean=501.8635642933046      std_dev=288.78251687655325
```

Serialized, all five agree: `mean 501.8636`, `std_dev 288.7825`.

Relative difference is around 1e-15, consistent with a different summation order.
Row-at-a-time accumulation in the incremental engine and batch accumulation in the
columnar and Arrow-backed engines do not add the same values in the same sequence.

## Why it is worth deciding

Bit-identical floating point across genuinely different accumulation strategies is
expensive to guarantee and would constrain future engine work, including the
Arrow-native paths. The realistic options:

1. Declare the serialized, rounded report the contract surface, and say so in
   `AGENTS.md` and the schema docs. Raw attributes carry full precision and are
   documented as accumulation-order dependent below the rounding precision.
2. Round at the accessor boundary as well, so attributes and serialization agree
   and cross-path equality holds everywhere. This is the strongest guarantee and
   the simplest to test, at the cost of discarding precision a caller might want.
3. Keep bit-identical accumulation as a requirement and add compensated summation
   (Kahan or Neumaier) to every engine. Most expensive, and still fragile.

Option 1 or 2 seems right. The choice matters beyond cosmetics because #526 is an
open bug where `to_llm_context` thresholds on unrounded values, so a value near a
threshold boundary can flip on the raw surface while the serialized report shows
the same rounded number. Settling this decides whether that class of bug is
possible at all.

- `AGENTS.md` states which surface the identical-numbers contract governs.
- The existing cross-engine parity tests assert against that surface explicitly,
  with a stated tolerance if raw attributes remain unrounded.
- Whatever is chosen is consistent with the resolution of #526.

Found during 0.10.0 dogfooding, sweeping sixteen input paths (CSV file/bytes/BytesIO
across four engines, JSON, JSONL, list[dict], Parquet file/bytes, pandas, polars,
pyarrow).

AndreaBozzo:
Closes #547.

## What this settles

The "profile numbers must be identical regardless of engine" rule was never
scoped, so it was read two ways: as a claim about raw `f64` values (which no
engine can guarantee, since accumulation order differs) or about what consumers
actually see. This scopes it to the second: **serialized, rounded metrics** —
Rust Serde output and Python `to_dict()` / `to_json()` / JSON `save()`.

- Native Rust fields and Python `ColumnProfile` attributes keep full precision
  and may differ in their final digits.
- Source and execution provenance (engine id, timings, memory, input bytes) is
  not an equality target.
- Absence is part of the contract: `None` means not analyzed, empty means
  analyzed and nothing found.
- A difference that survives serialization stays a parity defect, not an
  allowed raw-precision difference.

Documented in [the numeric contract](docs/schema/README.md#numeric-equality-contract),
with AGENTS.md, the Python guide and the docstrings pointing at it.

## Tests now enforce it

`test_engine_parity.py` and `tests/cross_engine_consistency.rs` compared a
hand-picked list of fields within a tolerance as loose as `0.01`. They now
compare the **complete serialized column documents** for exact equality, which
also covers optional statistics and absence markers — an absent statistic can
no longer silently compare equal to a measured one. Raw floats keep a separate
diagnostic check at rel `1e-9` / abs `1e-12`.

Coverage added: `csv.incremental`, `csv.columnar`, `csv.bytes`, `csv.buffer`,
`parquet.bytes`, and the sqlite arm folded onto the same assertion instead of
its own weaker loop.

## One real defect this found

Tightening the sqlite arm surfaced a genuine cross-path disagreement.
`analyze_column` fed raw values into the text statistics, so a `NULL` / `NaN`
token already counted as a null was *also* counted as a 4- or 3-character
string. The streaming engines feed a null-excluding length accumulator, so the
database connectors — the only other caller — disagreed with every file and
in-memory path:

| `null_str` = `["NULL","a","b","c","d"]` | min | max | avg |
| --- | --- | --- | --- |
| every file / in-memory engine | 1 | 1 | 1.0 |
| database connectors (before) | 1 | 4 | 1.6 |

Database profiles of string columns containing those tokens now report the same
lengths as every other path. This is the only reported-value change here; the
contract itself changes no rounding rule and no schema version.

## Verification

- `cargo test --workspace` — passes.
- `cargo fmt --all`, `cargo clippy --all --all-targets -- -D warnings` — clean
  on 1.98.1, matching the CI pin.
- `uv run pytest python/tests` — 1394 passed, 9 skipped (db-only).
- `uv run maturin develop --features "python,python-async,sqlite"` +
  `pytest python/tests/test_engine_parity.py` — 58 passed. The sqlite arm
  **fails on master's code** and passes with the fix; verified in both
  directions, and the new `text_lengths_exclude_null_like_tokens` unit test was
  confirmed to fail with the fix reverted alone.
- `uv run ruff format`, `ruff check`, `ty check` — clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Text length statistics now exclude null-like values such as `NULL` and `NaN`, ensuring consistent results across data sources.

* **Documentation**
  * Clarified numeric equality across engines and input formats for serialized reports.
  * Documented rounding, precision retention, sampling, missing values, and comparison behavior.
  * Added guidance for interpreting native floating-point values and loaded reports.

* **Tests**
  * Expanded cross-engine coverage and exact serialized-profile comparisons.
  * Added validation for precision preservation and consistent handling of text statistics.

## Review pass

Six findings from CodeRabbit and Copilot; five applied, one filed.

**Applied**

- *Cloning the column* (Copilot). The null-like filter built a second full
  `Vec<String>`, duplicating every string to drop a few. Added a borrowing
  variant of the text-stats entry point and collect `Vec<&String>` instead.
- *Frequency fields also change* (Copilot). Correct — the filtered slice also
  feeds `most_frequent`/`least_frequent`, so a null-like token no longer appears
  as a value and the remaining percentages are shares of the non-null values.
  The release note now says so.
- *"Not a change to reported values" is inaccurate* (Copilot). It is a change,
  for the database text statistics. The schema doc now claims only that no
  rounding rule and no schema version changed.
- *The threshold claim is too broad* (Copilot). Verified: `_dominant_pattern`
  compares native `confidence` against 0.5 while serialization rounds to 4dp, so
  a confidence just under the boundary is suppressed natively and emitted after
  a round trip. Narrowed the claim to the intent plus the known exception rather
  than changing threshold behaviour in this PR.
- *Quality is on the contract surface* (CodeRabbit). Added exact serialized
  quality assertions to both suites. It holds for every engine except pandas,
  which needs one bounded exception downstream of the int-to-float widening it
  already declares — pinned as a value rather than skipping the dimension, so
  nothing else can drift behind it.
- *State the analyzed-population condition* (CodeRabbit). Added to both Python
  docstrings.
- *The new round-trip test does not regress #547* (Copilot). Correct — it passes
  on the prior code, because nothing here was broken; what was missing was the
  statement of which surface equality is defined on. Reframed its docstring as a
  characterization rather than a regression guard. Did not add the suggested
  boundary assertion, because at the boundary the behaviour is the open defect
  above and the assertion would fail.

**Filed, not folded in: #709**

Copilot noticed that aligning the length statistics leaves `most_frequent` and
`least_frequent` still diverging. Confirmed on the same five values:

```
CSV engine:      {"Text":{"min_length":4,"max_length":5,"avg_length":4.6}}
analyze_column:  {"Text":{...,"most_frequent":[...],"least_frequent":[...]}}
```

`TextStats::from_lengths` hardcodes both to `None` and they are
`skip_serializing_if`, so the streaming builders serialize them absent while the
database path serializes them present. That is a live violation of the contract
this PR defines, and no test catches it: Python's `column_to_dict` drops both
fields, and the Rust suite compares two engines that share the `from_lengths`
builder. Resolving it means either dropping data Rust database consumers can
read today or computing top-N on a sample, so it is a behaviour decision rather
than something to fold in here. Recorded as a known defect in the schema doc
next to #637.

Gates re-run after these changes: `cargo test --workspace`, clippy on 1.98.1,
`pytest python/tests` (1394 passed), the sqlite arm (58 passed), ruff and ty.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -76,9 +76,15 @@ change; run what matches the change and list the commands in the PR body.
   into a plausible number is the worst bug class for a profiler.
 - **Semantics**: `None`/absent means "not analyzed"; empty means "analyzed,
   nothing found". Preserve this distinction in reports and bindings.
-- **Output contract**: profile numbers must be identical regardless of which
-  engine or input path produced them. If you touch an engine or parser, check
-  the other paths for parity.
+- **Output contract**: the serialized, rounded metric values must be identical
+  across engines and input paths for the same logical data, analysis options,
+  and analyzed population. Rust Serde output and Python `to_dict()`/`to_json()`
+  govern this contract; source/execution provenance is compared separately.
+  Native raw floats retain full precision and may differ with accumulation
+  order. Parity tests compare serialized values exactly; raw diagnostics use
+  explicit tolerances (relative `1e-9`, absolute `1e-12`). Preserve absence as
+  well as numbers. See [the numeric contract](docs/schema/README.md#numeric-equality-contract).
+  If you touch an engine or parser, check the other paths for parity.
 
 ## Project board rules
 
--- a/crates/dataprof-metrics/src/analysis/column.rs
+++ b/crates/dataprof-metrics/src/analysis/column.rs
@@ -160,7 +160,18 @@ fn analyze_column_with_options(
                     ColumnStats::None
                 }
             }
-            DataType::String | DataType::Identifier => calculate_text_stats(data),
+            DataType::String | DataType::Identifier => {
+                // Null-like tokens are counted as nulls above; they are not
+                // text values, so they must not enter the length statistics.
+                // The streaming engines feed a null-excluding accumulator here
+                // and this path has to agree with them (#547).
+                let values: Vec<String> = data
+                    .iter()
+                    .filter(|s| !is_null_like_token(s.trim()))
+                    .cloned()
+                    .collect();
+                calculate_text_stats(&values)
+            }
         }
     };
 
@@ -434,5 +445,27 @@ mod tests {
 
         assert_eq!(profile.null_count, 2); // 2 whitespace-only
         assert_eq!(profile.unique_count, Some(2)); // "value1" and "value2"
+    }
+
+    #[test]
+    fn text_lengths_exclude_null_like_tokens() {
+        // This path feeds the database connectors; the streaming engines use a
+        // null-excluding length accumulator. A token counted as a null must not
+        // also be counted as a four-character string (#547).
+        let data = vec![
+            "NULL".to_string(),
+            "a".to_string(),
+            "nan".to_string(),
+            "b".to_string(),
+        ];
+        let profile = analyze_column("test_col", &data);
+
+        assert_eq!(profile.null_count, 2);
+        let ColumnStats::Text(stats) = profile.stats else {
+            panic!("expected text stats");
+        };
+        assert_eq!(stats.min_length, 1);
+        assert_eq!(stats.max_length, 1);
+        assert_eq!(stats.avg_length, 1.0);
     }
 }
--- a/crates/dataprof-metrics/src/analysis/column.rs
+++ b/crates/dataprof-metrics/src/analysis/column.rs
@@ -7,7 +7,7 @@ use crate::analysis::inference::{
 };
 use crate::analysis::patterns::detect_patterns;
 use crate::stats::numeric::compute_numeric_stats_with_parsed_count;
-use crate::stats::{calculate_datetime_stats, calculate_text_stats};
+use crate::stats::{calculate_datetime_stats, calculate_text_stats_from_refs};
 
 /// Which parts of a column analysis to perform.
 ///
@@ -165,12 +165,11 @@ fn analyze_column_with_options(
                 // text values, so they must not enter the length statistics.
                 // The streaming engines feed a null-excluding accumulator here
                 // and this path has to agree with them (#547).
-                let values: Vec<String> = data
+                let values: Vec<&String> = data
                     .iter()
                     .filter(|s| !is_null_like_token(s.trim()))
-                    .cloned()
                     .collect();
-                calculate_text_stats(&values)
+                calculate_text_stats_from_refs(&values)
             }
         }
     };
--- a/crates/dataprof-metrics/src/stats/mod.rs
+++ b/crates/dataprof-metrics/src/stats/mod.rs
@@ -8,4 +8,4 @@ pub use accumulator::NumericAccumulator;
 pub use cardinality::{CardinalityEstimator, EXACT_CARDINALITY_THRESHOLD, HyperLogLog};
 pub use datetime::calculate_datetime_stats;
 pub use numeric::calculate_numeric_stats;
-pub use text::calculate_text_stats;
+pub use text::{calculate_text_stats, calculate_text_stats_from_refs};
--- a/crates/dataprof-metrics/src/stats/text.rs
+++ b/crates/dataprof-metrics/src/stats/text.rs
@@ -9,9 +9,27 @@ pub fn calculate_text_stats(data: &[String]) -> ColumnStats {
     ColumnStats::Text(compute_text_stats(data))
 }
 
+/// Borrowing variant for callers that must filter the column first.
+///
+/// Filtering through references keeps the caller from duplicating every string
+/// in the column just to drop a few of them.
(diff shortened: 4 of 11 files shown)


AndreaBozzo:
Thanks — both reviews found real things. Five applied, one filed.

**Applied**

- The `Vec<String>` clone is gone; there is now a borrowing variant of the
  text-stats entry point and the filter collects `Vec<&String>`.
- The frequency fields do change too, and the release note now says so
  explicitly rather than claiming lengths were the only effect.
- "Not a change to reported values" was wrong; the schema doc now claims only
  that no rounding rule and no schema version changed.
- The `to_llm_context()` threshold claim was too broad. Verified against
  `_columns.py`: `_dominant_pattern` compares native `confidence` against 0.5
  while serialization rounds to four decimals, so a value just under the
  boundary is suppressed natively and emitted after a round trip. Narrowed the
  claim rather than changing threshold behaviour in this PR.
- Quality is on the contract surface — added exact serialized quality
  assertions to both suites. Holds for every engine except pandas, which needs
  one bounded exception downstream of the int-to-float widening it already
  declares. Pinned as a value rather than skipping the dimension so nothing
  else can drift behind it.
- Added the analyzed-population condition to both Python docstrings.
- On the round-trip test not regressing #547: correct, it passes on the prior
  code. Reframed its docstring as a characterization rather than a regression
  guard. I did not add the suggested boundary assertion, because at the
  boundary the behaviour is the open defect above and the assertion would fail.

**Filed as #709**

The observation that `most_frequent`/`least_frequent` still diverge is right,
and it is the most useful finding here. Reproduced on the same five values:

```
CSV engine:      {"Text":{"min_length":4,"max_length":5,"avg_length":4.6}}
analyze_column:  {"Text":{...,"most_frequent":[...],"least_frequent":[...]}}
```

`TextStats::from_lengths` hardcodes both to `None` and they are
`skip_serializing_if`, so the streaming builders serialize them absent while
the database path serializes them present — a live violation of the contract
this PR defines, invisible to both suites today. Fixing it means either
dropping data Rust database consumers can read or computing top-N on a sample,
so it is a behaviour decision rather than something to fold in here. Recorded
as a known defect in the schema doc next to #637.

Issue #710, opened 2026-09-10
python -m dataprof.check: a CI entrypoint over the quality-gate API


AndreaBozzo:
Parent: #371

## User problem

Running a quality gate in CI currently means writing a Python script: import
dataprof, profile the file, load a baseline, decide which fields to compare,
invent a verdict, and translate it into an exit code. Every adopter writes that
script, and each one invents different missing-metric handling and a different
notion of failure.

`examples/etl_quality_gate.rs` shows the shape of the problem from the Rust
side: the gate is a `struct Gate` declared inside the example, because there is
nowhere else for it to live.

#376 gives the decision a library API. This issue gives it an invocation that a
CI job, a Makefile, or an orchestrator task can call in one line without a
wrapper script.

## Proposed contract

A module entrypoint over the #376 API, with no new profiling semantics:

```bash
python -m dataprof.check data.csv --min-quality 90 --baseline profile.json
```

- Exits `0` when the policy passes, non-zero when it fails, and a distinct
  non-zero code when the policy could not be evaluated (missing metric, source
  not exhausted, unreadable baseline).
- Writes the structured `CheckResult` from #376 to stdout as JSON when asked,
  so a job can archive the verdict alongside the run.
- Reads the policy from flags or from a config file, with the file as the form
  that survives review.
- Prints a human-readable summary to stderr, keeping stdout clean for machine
  consumption.

This is a thin argument-parsing and exit-code layer. Every decision it makes
belongs to #376; if a behavior cannot be expressed through the library API, it
does not belong here.

## Relationship to the no-CLI-binary rule

The release surface stays Rust crates plus Python wheels. This adds no compiled
binary, no `[[bin]]` target, and no `console_scripts` entry point. It is a
`__main__` module inside the existing wheel, the same shape #330 proposes for
the agent surface, reachable only through the interpreter that already has
dataprof installed.

## Acceptance criteria

- [ ] `python -m dataprof.check --help` documents every flag and every exit code.
- [ ] Exit codes distinguish pass, fail, and not-evaluable; the last is never
      collapsed into either of the others.
- [ ] `None` (not analyzed) never silently satisfies or violates a threshold.
- [ ] Machine output on stdout is valid JSON and contains nothing else.
- [ ] The entrypoint adds no policy logic that #376 does not expose.
- [ ] Tests cover pass, fail, missing metric, absent baseline, and malformed
      policy file.
- [ ] Documented in the Python guide with a copy-pasteable CI job.

[...]

PR #725, opened 2026-09-13
feat: add a batch quality-gate API with structured results

### LINKED ISSUE #376: Add a small batch quality-gate API with structured results
Parent: #371

## User problem

The README shows users how to inspect a score, but every CI or ETL user must hand-roll comparisons, missing-metric handling, messages, and exit behavior. #188 concerns continuous streaming alerts; batch profiles still need a small library-level decision API.

## Proposed shape

Provide an explicit policy/result model, for example:

```python
result = report.check(
    min_quality=90,
    max_null_percentage={"customer_id": 0, "*": 20},
    require_metrics=["quality"],
)
if not result.passed:
    for violation in result.violations:
        print(violation.code, violation.message)
```

The exact names should be settled before implementation. The important contract is that evaluation returns structured deterministic data and does not exit the process, print, send alerts, or mutate the report.

## Acceptance criteria

- [ ] Supports an overall quality floor and at least one column-level threshold.
- [ ] Missing/not-analyzed metrics produce an explicit policy outcome; they never pass as zero or fail accidentally due to a default.
- [ ] Violations have stable codes, observed value, expected constraint, and optional column.
- [ ] Results serialize cleanly for CI systems.
- [ ] Rust and Python semantics match for the shared subset.
- [ ] README/example coverage shows an ETL gate without a CLI.
- [ ] Tests cover pass, fail, missing metric, wildcard/column override, and percentage scale (0–100).

## Depends on

- **#715**, hard. This ticket requires that a missing or not-analyzed metric
  produce an explicit policy outcome and never pass as zero or fail by default.
  That is not implementable today: `ReportAssembler::build()` cannot fail, so a
  quality computation that was never requested and one that failed internally
  both arrive as `quality = None`. A gate built on top would give the same
  verdict for "you did not ask for this" and "this broke".

## Relationship to other work

- May consume the structured finding/evidence vocabulary from the findings child of #371, but must not require findings to evaluate numeric policies.
- Does not replace #188, which owns continuous stream triggers and sinks.
- Must preserve `None` = not analyzed versus empty = analyzed with no violations.

## Likely files

- `crates/dataprof-core/src/`
- `crates/dataprof-runtime/src/`
- `python/dataprof/__init__.py` and stubs
- `examples/etl_quality_gate.rs` and Python equivalent
- focused Rust/Python tests

## Non-goals

- process exit codes or CLI flags
- webhooks, logging sinks, or schedulers
- transforming data to make a gate pass

## Evidence scope and incomplete evaluation

A decision needs both its result and the evidence scope supporting it. Current execution metadata already exposes source exhaustion, truncation, and sampling; this API should use that information when a policy asks about the whole source.

- [ ] State whether the decision concerns the scanned rows/selected columns or the full source, and preserve the reason evidence is insufficient.
- [ ] A clean prefix or sample cannot pass a full-source uniqueness requirement. A witnessed duplicate may establish failure for that requirement despite an incomplete scan.
- [ ] Do not generalize sample ratios into full-source pass/fail outcomes without valid bounds. Scope-specific policies can still evaluate the observed sample explicitly.
- [ ] Separate an unrequested or unsupported metric from a completed check with no violations; a requested whole-row check must not appear evaluated after column projection.
- [ ] Cover complete input, capped input, unavailable metrics, projected columns, and a conclusive violation with incomplete coverage in matching Rust/Python fixtures.
- [ ] Preserve verdict, scope, and evidence status/reason when serializing. Keep numeric gate evaluation independent of findings; coordinate vocabulary with #375.

The design motivation is the separation of a verdict from evidence completeness in [mission-level runtime assurance, section 4](https://arxiv.org/html/2607.23532v2). Applying it to profiler policies is a design inference, not a guarantee established by that paper. This does not require a distributed engine or a new global report schema.

AndreaBozzo:
Closes #376.

## Why this one matters

Until now dataprof has been something a person runs. You point it at a file,
you read the report, you learn something, and then the tool is done. That caps
its value at one run per dataset, and it means the thing dataprof is best at
(noticing that data stopped looking like itself) only happens when somebody
remembers to look.

A gate changes who does the running. Once a policy is written down, the
profiler executes on every load, every PR, every scheduled job, and nobody
decides to invoke it. The second run is worth more than the first, because by
then there is something to compare against and something to refuse.

That is the shift this PR starts: from the first ten minutes with unfamiliar
data to a checkpoint that stays in the pipeline. #710 (`python -m
dataprof.check`) and the Marketplace action build directly on what lands here.

## The problem with hand-rolled gates

Both `etl_quality_gate` examples used to be about 150 lines of hand-written
checks, and that was not because the checks are hard. It was because every
caller has to re-decide the same four questions, in the same order, every time:

1. **What does a missing metric mean?** `report.quality` is `None` when the
   quality pack was deselected, when the source held nothing to measure, when
   projection withheld the dimensions, and when the computation failed. A gate
   that reads absence as "skip" passes a broken run. A gate that reads it as
   zero fails a correctly configured one.
2. **What does a number over part of the source prove?** A 2% null rate
   measured on the first 10,000 rows says nothing about row 10,001. Most gates
   treat it as if it did.
3. **Which value goes in the log line?** Enough for an on-call engineer to
   triage without re-running the profiler.
4. **Is the threshold a percentage or a ratio?** Writing `0.9` when the scale
   is `0..100` produces a gate that silently never fires, which is the worst
   possible failure mode for a check.

Getting those wrong is quiet. The pipeline stays green and the gate guards
nothing. So the answers belong in the library, decided once.

## What the API looks like

```python
result = report.check(
    min_quality_score=90,
    max_null_percentage={"customer_id": 0, "*": 20},
    max_duplicate_rows=0,
    require_metrics=["quality"],
)

if not result.passed:
    for check in result.violations:
        print(check.code, check.column, check.message, check.observed)
```

```rust
let result = QualityPolicy::new()
    .min_quality_score(90.0)
    .max_null_percentage("customer_id", 0.0)
    .max_null_percentage_any(20.0)
    .max_duplicate_rows(0)
    .require_quality()
    .evaluate(&report)?;
```

Evaluation returns data. It does not print, does not exit the process, and does
not touch the report, so it drops into Airflow, Dagster, a GitHub Action, or a
shell script without a CLI existing.

## The idea worth arguing about: three verdicts

A gate normally answers yes or no. This one answers three ways, and the third
is the point.

| Verdict | Meaning |
|---|---|
| `fail` | A requirement was conclusively violated |
| `pass` | Every requirement was evaluated and met |
| `inconclusive` | Nothing was violated, and something could not be checked |

`inconclusive` covers an unanalyzed metric, a column the report has no profile
for, or evidence that does not reach as far as the requirement does.
`result.passed` is true only for `pass`, so an unanswerable gate never reads as
a green one, and a caller who wants to distinguish "bad data" from "cannot
tell" reads `result.verdict` instead.

The reason to separate them: those two situations need different humans. A
violation is a data problem and belongs to whoever produced the feed. An
inconclusive result is usually a configuration problem, and belongs to whoever
set up the gate. Collapsing them into one red light sends every alert to the
wrong person half the time.

`fail` wins over `inconclusive`, because a violation that was actually
witnessed is a decision regardless of what else went unchecked.

## The idea behind that: a verdict and its evidence are separate facts

Every policy declares what it is a statement about. The default,
`scope="full_source"`, means the requirements are claims about the entire
dataset. Under that scope:

- **A ratio computed over part of a source generalizes in neither
  direction.** A clean first 10,000 rows do not make a clean file, and a dirty
  sample does not prove a dirty file either. Such a check is left unevaluated,
  with the observed number still attached so it is readable.
- **A witnessed count does generalize, one way.** Rows already seen as
  duplicates stay duplicates when more rows are read, so an exact duplicate
  count above the allowance fails conclusively even on a capped scan. A count
  at or below the allowance is still not a pass.
- **An estimated count witnesses nothing.** Once the duplicate counter has
  spilled to its HLL sketch, the number is not a row anyone saw, and it settles
  neither direction.

`scope="observed"` states the requirements about whatever the metrics actually
measured, which is always evaluable and makes no claim about unread rows. It is
the right scope for "was the sample we took acceptable", and the wrong one for
"can we load this".

The design motivation is the separation of a verdict from evidence
completeness in [mission-level runtime assurance, section
4](https://arxiv.org/html/2607.23532v2). Applying it to profiler policies is a
design inference, not a guarantee that paper establishes.

## The bug this would have shipped with, and how it was caught

The obvious implementation reads evidence out of `ExecutionMetadata`:
`source_exhausted`, `truncation_reason`, `sampling_applied`. The ticket says as
much, and it is wrong on exactly the files where a gate matters.

Quality metrics are computed from a bounded reservoir of 10,000 values per
column. On a 12,000-row CSV read start to finish with no row sampler,
`source_exhausted` is `true` and `sampling_applied` is `false`, both correctly,
and the quality score still came from a sample. A gate trusting execution
metadata alone would certify that score as full-source evidence.

`MetricConfidence` already records this in Rust, and the Python layer had no
access to it. So this PR adds `QualityAssessment::sampled_dimensions()` and
exposes it as `report.quality_sampled_dimensions`, with the Python report
document recording it alongside the metrics. Provenance is resolved per
component, so a completeness score from exact column counters stays decidable
while a consistency score from the reservoir does not.

Worth recording how this surfaced: stubbing the accessor to return "nothing was
sampled" passed the entire test suite, because every fixture was small enough
that the reservoir held everything. The tests were guarding one step short of
the failing path. `test_a_fully_read_large_file_still_has_a_sampled_quality_score`
and the two reservoir tests in `quality_gate.rs` now reach it, and they were
each confirmed to fail against the stub.

## Cross-language parity

Rust and Python have separate implementations, so a gate that disagreed between
them would let the same dataset pass in one pipeline and fail in the next.
`tests/fixtures/quality_gate_parity.json` states one CSV, the profiling
options, the policy, and the exact expected result document for eight cases.
`tests/quality_gate_parity.rs` asserts the Rust evaluator against it and
`python/tests/test_quality_gate_parity.py` asserts the Python one, so changing
either implementation alone fails that layer's test. Perturbing one expected
verdict in the fixture was confirmed to fail both.

The cases cover a complete input, a violated threshold, an unavailable metric
both with and without `require_metrics`, projected columns, a capped input
under both scopes, and a violation witnessed under incomplete coverage.

Two details that keep the two documents identical:

- **Messages carry no numbers.** `observed` and `expected` hold the values, so
  nothing depends on how a float was formatted into a sentence. Rust's `{:.2}`
  rounds half away from zero and Python's `format` rounds half to even; a
  single number inside a string would have been a latent parity bug.
- **Evaluation order is fixed by the data, not the policy.** Named column
  limits evaluate in column-name order and wildcard limits in report column
  order, so a policy read out of a JSON object evaluates identically in both
  languages regardless of key ordering.

## Deliberately out of scope

**Column-presence requirements.** A column absent from a report was either
projected away or not in the source, and a report does not record which. Rather
than guess, a named column with no profile is reported as
`column_not_profiled` and left undecided. Both examples show the two-line
schema check alongside the policy, which is honest about the run actually
profiling everything. Recording projection in the report would make this
checkable and is worth its own ticket.

Also out of scope, per the ticket: process exit codes, CLI flags, webhooks,
schedulers, and anything that transforms data to make a gate pass.

## Behaviour changes

- `ProfileReport.to_dict()` gains `quality.sampled_dimensions` when the report
  carries an assessment. Additive, `schema_version` unchanged, and the
  committed JSON schema is regenerated. Absence is preserved on reload: a
  document written before this field does not say how its quality numbers were
  obtained, so a full-source policy over it reports `coverage_unrecorded` and
  comes back inconclusive rather than assuming an exact scan.
- Both `etl_quality_gate` examples are rewritten on the API and gain a fourth
  drop, a capped scan over a file whose duplicate sits past the cap, so the
  `inconclusive` verdict appears in real output rather than only in prose.
- New public names: `dataprof.QualityCheck`, `dataprof.QualityGateResult`,
  `ProfileReport.check()`, `ProfileReport.quality_sampled_dimensions` in
  Python; `QualityPolicy`, `GateResult`, `Check`, `Verdict`, `PolicyScope`,
  `Evidence`, `EvidenceGap`, `CheckCode`, `CheckStatus`, `Expectation`,
  `MetricValue`, `NotEvaluated`, `PolicyError`, `RequiredMetric` in Rust.

## Verification

```
cargo fmt --all
cargo test --workspace
RUSTUP_TOOLCHAIN=1.96 python .github/scripts/msrv_check.py
uv run ty check python/
uv run pytest python/tests/ -q        # 1500 passed, 9 skipped
cargo run --example etl_quality_gate
uv run python python/examples/etl_quality_gate.py
```

Each new guard was run against the unfixed code and confirmed to fail:
removing the evidence gate (separately in Rust and in Python), stubbing the
provenance accessor, reverting the duplicate-column-name lookup, and perturbing
one expected verdict in the shared fixture.

## Summary by CodeRabbit

- **New Features**
  - Added declarative quality gates with configurable quality, null, duplicate, dimension, and metric requirements.
  - Gate results now report pass, fail, or inconclusive verdicts with per-check evidence and reasons.
  - Added Python `ProfileReport.check()` support and public quality-gate result types.
  - Added visibility into quality metrics calculated from retained samples.
  - Added full-source and observed-scope evaluation options.

- **Documentation**
  - Expanded guides, schemas, and examples for quality gates, sampling coverage, and serialized results.

- **Tests**
  - Added cross-language parity coverage for gate evaluation and verdict behavior.


AndreaBozzo:
Addressed in 7f99adc, except one.

**Fixed**

- **Legacy provenance** (both reviewers, `quality_gate.rs`/`quality.rs`/`profile_report.rs`): confirmed. `deserialize_quality_compat` claimed `Exact` for a flat legacy quality object, which the document never stated, and the Python reload path of the same report said `coverage_unrecorded`. Added `MetricConfidence::Unrecorded`; `sampled_dimensions()` now returns `Option<Vec<String>>` with `None` for it, and both layers report `coverage_unrecorded`. Guarded by `a_legacy_document_without_recorded_coverage_is_not_read_as_a_full_scan`, confirmed to fail against the old `QualityAssessment::exact` path.

- **Zero-weight dimension taints the overall score** (Copilot, both languages): confirmed. Overall provenance is now the sampled components belonging to `assessed_dimensions()`, which is exactly the weighted contributing set and is available identically to dict-backed reports. Guarded on both sides.

- **Zero weights erase the sampling record** (CodeRabbit, outside-diff): confirmed. `QualityAssessment::new` downgraded to `NotAssessed` based on the *weighted* set, so a measured but unweighted dimension lost its provenance entirely. The downgrade is now weight-independent: it fires only when no dimension has a score at all.

- **Examples**: acceptance now requires `missing_columns` to be empty, findings chain violations with unevaluated checks, and the stale "a capped scan still sees it" comment is gone (the capped scan is the fourth drop and does not see the duplicate).

- **Parity guard** now pins the passing observed-scope case, and **the fixture** gains a source read in full whose quality came from the bounded reservoir. That was the real gap: every previous incomplete-evidence case reached its verdict through truncation, so the two implementations could have disagreed on sampled-dimension handling while parity stayed green. Stubbing Python sampled-evidence to "complete" now fails the parity test.

**Not taking**

- **Guard `duplicate_rows` for dict-backed reports** (CodeRabbit, `_gate.py`): declining. A uniqueness object with `rows_checked` but no `duplicate_rows` is malformed, and Rust rejects such a document at decode time rather than gating it (`UniquenessMetrics.duplicate_rows` has no serde default). Returning `not_assessed` would make Python answer where Rust refuses to load, and it would turn a decode problem into a plausible-looking verdict, which this repo explicitly avoids on decode paths. The `KeyError` is loud and correct; if that document shape should be handled, it belongs in `from_dict` validation, not in the gate.

Each fix was reverted individually and confirmed to fail its own guard.

